PR #126, opened 2020-10-27
release tag

asoorm:


asoorm:
Closing because goreleaser is being used instead when we push tags.

PR #131, opened 2020-10-29
Update installation.md

rewsmith:
Included an instruction to create the tyk-operator-system ns, and added -n cert-manager on k get all command


asoorm (approved):
Thank you @rewsmith

PR #132, opened 2020-10-29
add support for  internal looping API

### LINKED ISSUE #133: feat: url rewrite to internal APIs
Dynamic Auth

In this example, we have a bunch of legacy customers who authenticate with our service using Basic Authentication.
We want to be able to support API Keys also, where both types of clients hit the same ingress.
As such, Tyk needs to decide whether to perform Basic Authentication or Auth Token auth check.

In order to achieve this, we need to configure 4 API Definitions inside Tyk.

1. EntryPoint API
2. BasicAuthInternal API
3. AuthTokenInternal API
4. ProxyInternal API

When the request hits the ingress route, we configure a URL rewrite to pass the request to either the internal
BasicAuth api or the AuthToken API.
the internal APIs will then authenticate request, and assuming the happy path, proxy to the ProxyInternal API.
The ProxyInternal API is responsible for proxying to the underlying service.

It is worth noting that there are no actual http redirect happening here, meaning that there is no performance penalty
in performing any of these "Internal Redirects".

Rather than having to specify the APIID, to proxy to, we should be able to target the namespace/name of the APIDefinition custom Resource.

If the BasicAuthInternal or ProxyInternal custom resources are attempted to be deleted, then the finalizer should fail & retry till the API(s) which are referencing it are either also deleted or de-reference.

```yaml
---
kind: ApiDefinition
metadata:
  name: entrypoint-api
spec:
  name: Entrypoint API
  protocol: http
  active: true
  proxy:
    listen_path: /entrypoint
    strip_listen_path: true
    # TODO: We should be able to create ROUTES without a target URL.
    target_url: http://dummy.com
  use_keyless: true
  # if Authorization: Basic - redirect to tyk://APIID where APIID is the ID associated with namespace/name e.g. default/basic-auth-internal
  # if Authorization: Bearer - redirect to tyk://APIID where APIID is the ID associated with namespace/name e.g. default/auth-token-internal
---
kind: ApiDefinition
metadata:
  name: basic-auth-internal
spec:
  name: BasicAuth Internal API
  protocol: http
  active: true
  internal: true
  use_basic_auth: true
  auth_configs:
    basicAuth:
      auth_header_name: Authorization
  version_data:
    default_version: Default
    not_versioned: true
    versions:
      Default:
        name: Default
        # redirect all traffic to proxy API
---
kind: ApiDefinition
metadata:
  name: auth-token-internal
spec:
  name: AuthToken Internal API
  protocol: http
  active: true
  internal: true
  use_basic_auth: true
  auth_configs:
    authToken:
      auth_header_name: Authorization
  version_data:
    default_version: Default
    not_versioned: true
    versions:
      Default:
        name: Default
        # redirect all traffic to proxy API
---
kind: ApiDefinition
metadata:
  name: proxy-api
spec:
  name: Proxy API
  protocol: http
  active: true
  internal: true
  proxy:
    target_url: http://httpbin.org
  use_keyless: true
---
```

@buger FYI

asoorm:
Closes #133

TODO: 

- [x] implement Looping / URL Rewrite middleware.
- [ ] Update documentation on the feature
- [ ] add bdd tests


jjensen90:
How's this looking lately @asoorm? Did you still plan to use this PoC PR to enable/document looping?

api/v1alpha1/apidefinition_types.go
```diff
@@ -163,8 +168,32 @@ type URLRewriteMeta struct {
...
 	// RewriteTo is the target path on the upstream, or target URL we wish to rewrite to
-	RewriteTo string           `json:"rewrite_to"`
-	Triggers  []RoutingTrigger `json:"triggers"`
+	// +optionall
+	RewriteTo string `json:"rewrite_to,omitempty"`
+	// RewriteToLoop serves as rewrite_to but used when rewriting to looping
+	// targets. When rewrite_to and rewrite_to_loop are both provided then
+	// rewrite_to will take rewrite_to_loop
+	// +optionall
+	RewriteToLoop *LoopTarget `json:"rewrite_to_loop,omitempty"`
```
asoorm, line 177: The API Definition specifies `internal: true`.
So rather than calling it `RewriteToLoop`, we should call it `RewriteInternalTarget` or similar?
Further - how about `namespace/name`?

I want to rewrite to a target in a different k8s namespace? Do we want to enable this capability?

asoorm:
> How's this looking lately @asoorm? Did you still plan to use this PoC PR to enable/document looping?

We are close @jjensen90 - Hopefully this feature should be available within the next week(s).

api/v1alpha1/apidefinition_types.go
```diff
@@ -163,8 +170,34 @@ type URLRewriteMeta struct {
...
+}
+
+// LoopInternal defines options that constructs a url that refers to an api that
+// is loaded into the gateway.
+type LoopInternal struct {
+	// Self set this to true if the api definition is looping to itself.
+	Self bool `json:"self,omitempty"`
+	// API a namespaced/name to the api definition resource that you are
+	// targetting. If self is set to true this value will be ignored.
+	//	example
```
asoorm, line 192: For MVP - Let's not include this, until we have concrete use-case / example?

api/v1alpha1/apidefinition_types.go
```diff
@@ -163,8 +170,34 @@ type URLRewriteMeta struct {
...
+	RewriteToInternal *LoopInternal `json:"rewrite_to_internal,omitempty"`
+	// +optionall
+	Triggers []RoutingTrigger `json:"triggers,omitempty"`
+}
+
+// LoopInternal defines options that constructs a url that refers to an api that
+// is loaded into the gateway.
+type LoopInternal struct {
+	// Self set this to true if the api definition is looping to itself.
+	Self bool `json:"self,omitempty"`
```
asoorm, line 189: For MVP - Let's not include `Self`, until we have concrete use-case / example and can create a BDD feature which clearly describes and tests that use-case?

api/v1alpha1/apidefinition_types.go
```diff
@@ -163,8 +170,34 @@ type URLRewriteMeta struct {
...
+	// API a namespaced/name to the api definition resource that you are
+	// targetting. If self is set to true this value will be ignored.
+	//	example
+	API string `json:"api,omitempty"`
```
asoorm, line 193: As per ticket update, check whether we should be able to route to an internal API in another namespace?

How about referencing the API Definition as:

```
target:
  namespace: foo
  name: bar
```

If the namespace is omitted - then reference the namespace of current object.

And if we don't need the namespace because we can only link to resources in the current namespace, we could just define the target as `target: string` to reference the name of target.

api/v1alpha1/apidefinition_types.go
```diff
@@ -137,23 +137,30 @@ type CircuitBreakerMeta struct {
 
 type StringRegexMap struct {
 	MatchPattern string `json:"match_rx"`
-	Reverse      bool   `json:"reverse"`
-	//matchRegex   *regexp.Regexp
+	// +optionall
```
asoorm, line 140: reminder - delete these `// +optionall` also spelling issue

api/v1alpha1/apidefinition_types.go
gernest: done bc8072c

api/v1alpha1/apidefinition_types.go
gernest: done 33def54

api/v1alpha1/apidefinition_types.go
gernest: done 5e73c71

api/v1alpha1/apidefinition_types.go
gernest: done 33def54

api/v1alpha1/apidefinition_types.go
```diff
@@ -593,7 +619,8 @@ type Proxy struct {
 	ListenPath string `json:"listen_path,omitempty"`
 
 	// TargetURL defines the target URL that the request should be proxied to.
-	TargetURL string `json:"target_url"`
+	TargetURL     string        `json:"target_url,omitempty"`
+	TargeInternal *LoopInternal `json:"target_internal,omitempty"`
```
asoorm, line 623: spelling: TargeInternal -> TargetInternal

config/crd/bases/tyk.tyk.io_apidefinitions.yaml
```diff
@@ -1425,31 +1474,48 @@ spec:
...
-                                        - path_part_matches
-                                        - payload_matches
-                                        - query_val_matches
-                                        - request_context_matches
-                                        - session_meta_matches
                                         type: object
                                       rewrite_to:
                                         type: string
+                                      rewrite_to_internal:
+                                        description: LoopInternal defines options
```
asoorm, line 1483: LoopInternal -> RewriteToInternal?

controllers/apidefinition_controller.go
```diff
@@ -285,6 +288,47 @@ func (r *ApiDefinitionReconciler) updateLinkedPolicies(ctx context.Context, a *t
...
+					if x.RewriteToInternal != nil {
+						x.RewriteTo = formatLoop(x.RewriteToInternal)
+						x.RewriteToInternal = nil
+					}
+				}
+			}
+		}
+		d.Versions[n] = v
+	}
+}
```
asoorm, line 317: Would be great to see some tests - specifically for this loop.

gernest: done f6c706b

config/crd/bases/tyk.tyk.io_apidefinitions.yaml
gernest: done d369a51

api/v1alpha1/apidefinition_types.go
gernest: done d369a51

PR #139, opened 2020-10-30
feat: add coprocess auth

Jesse0Michael:
Enable the `enable_coprocess_auth` and `custom_middleware.id_extractor.extractor_config` api definition fields to allow configuring co-process auth through the Tyk operator.

Pulled extractor config values from https://tyk.io/docs/plugins/auth-plugins/id-extractor/ and https://github.com/TykTechnologies/tyk/blob/master/gateway/coprocess_id_extractor.go to not miss any options and satisfy the generator (which didn't want any interface{} types)

---

I tested this change locally with a tyk-operator build that included these changes and a local tyk gateway (community edition) using an ApiDefinition CRD that included these values
``` yaml
  enable_coprocess_auth: true
  custom_middleware:
    driver: grpc 
    auth_check:
      name: TestAuth
      path: ""
      require_session: false
    id_extractor:
      extract_from: header
      extract_with: value
      extractor_config:
        header_name: X-Test-Token
```

Please let me know any feedback on how enabling this feature should be implemented, tested, or documented.


asoorm:
Thank you @Jesse0Michael ! I need to find a nice way of running these tests from forks with GH Secrets.

asoorm (approved):

PR #159, opened 2020-11-09
dev: instruction for local development environment

gernest:
I am setting up my local development env. This documents the journey.

Nothing to see for now. If you are interested, every commit  documents what changed. 

- create/destroy a local `kind` cluster
- boostrap ce/pro tyk
- deploy operator to local registry
- faster image build for local dev
- use `go test ./bdd/`  avoid installing extra tool for testing


asoorm:
Thanks @gernest 

Now we have 2 dev env setup docs? https://github.com/TykTechnologies/tyk-operator/blob/master/docs/development.md
Can we sort out a Zoom to go through this?

PR #164, opened 2020-11-13
ability to restrict watched resources by namespace

asoorm:
Allows tyk-operator to watch `ALL`, `ONE` or `SPECIFIC` namespaces.

e.g. `team-a`, `team-b`, `team-c`, `team-d` might exist. But if the operator is only watching the `team-a` namespace, then APIDefinition and other Tyk custom resources will have 0 effect if they are outside of the operator's scope.

This facilitates having multiple namespace scoped operators deployed inside a single cluster.
By inference, multiple Tyk deployments or organisations inside a single cluster. We shall assume that a single operator operates on a single Tyk Organization.

### DIFF
--- a/main.go
+++ b/main.go
@@ -19,9 +19,12 @@ package main
 import (
 	"errors"
 	"flag"
+	"fmt"
 	"os"
 	"strconv"
 	"strings"
+
+	"sigs.k8s.io/controller-runtime/pkg/cache"
 
 	"github.com/TykTechnologies/tyk-operator/pkg/dashboard_admin_client"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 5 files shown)


gernest (changes requested):
I haven't tested yet , just left a comment

So I have following namespaces

```
NAME                  STATUS   AGE
cert-manager          Active   2d22h
default               Active   2d22h
kube-node-lease       Active   2d22h
kube-public           Active   2d22h
kube-system           Active   2d22h
local-path-storage    Active   2d22h
tyk-operator-system   Active   2d22h
tykce-control-plane   Active   2d22h
```

Creating apidef in `default` namespace I get this error inside the controller

```
... "httpbin.164767f486037709" is forbidden: User "system:serviceaccount:tyk-operator-system:default" cannot patch resource "events" in API group "" in the namespace "default"' (will not retry!)
```

Does this PR address this ? What was the previous behaviour?

asoorm:
> I haven't tested yet , just left a comment
> 
> So I have following namespaces
> 
> ```
> NAME                  STATUS   AGE
> cert-manager          Active   2d22h
> default               Active   2d22h
> kube-node-lease       Active   2d22h
> kube-public           Active   2d22h
> kube-system           Active   2d22h
> local-path-storage    Active   2d22h
> tyk-operator-system   Active   2d22h
> tykce-control-plane   Active   2d22h
> ```
> 
> Creating apidef in `default` namespace I get this error inside the controller
> 
> ```
> ... "httpbin.164767f486037709" is forbidden: User "system:serviceaccount:tyk-operator-system:default" cannot patch resource "events" in API group "" in the namespace "default"' (will not retry!)
> ```
> 
> Does this PR address this ? What was the previous behaviour?

I think your issue is different. Check master:

https://github.com/TykTechnologies/tyk-operator/blob/master/Makefile#L66-L67

gernest (approved):

main.go
```diff
@@ -174,13 +174,18 @@ func main() {
...
 // which specifies the Namespace to watch.
 // An empty value means the operator is running with cluster scope.
 const watchNamespaceEnvVar = "WATCH_NAMESPACE"
 
 // getWatchNamespace returns the Namespace the operator should be watching for changes
 func getWatchNamespace() (string, bool) {
+	// WatchNamespaceEnvVar is the constant for env variable WATCH_NAMESPACE
+	// which specifies the Namespace to watch.
+	// An empty value means the operator is running with cluster scope.
+	const watchNamespaceEnvVar = "WATCH_NAMESPACE"
```
gernest, line 188: @asoorm  This is only used within the function call, used only once, and can't be changed without looking at the function call (Meaning not meant to easy refactoring). The doc comment looks like a global exported symbol to me.

In this case `os.LookupEnv( "WATCH_NAMESPACE")`  should be fine

Did I miss something?

Issue #171, opened 2020-11-24
docs: how to configure JetBrains & VS code intellisense & validation plugins


asoorm:
Provide nav heading such as `IDE Support`. Pointing to JetBrains & VS Code Kubernetes plugins.

Example video showing VS Code intellisense & validation of custom resources.

VS Code: https://marketplace.visualstudio.com/items?itemName=ms-kubernetes-tools.vscode-kubernetes-tools
JetBrains: https://plugins.jetbrains.com/plugin/10485-kubernetes

![Screenshot 2020-11-24 at 13 37 09](https://user-images.githubusercontent.com/1465130/100101606-5f8de080-2e5a-11eb-94d6-8f0ded7372c8.png)

PR #178, opened 2020-11-26
Added IDE Integration section with VS Code support

hellobudha:
A new section detailing the steps required to add K8s extensions to popular IDEs.
Added instructions to install K8s extension for VS Code.


asoorm:
resolves #171

PR #192, opened 2020-12-02
Update helm chart when a version is pushed

### LINKED ISSUE #191: feat: cut a release
https://github.com/TykTechnologies/tyk-operator/actions?query=workflow%3A%22Cut+Release%22

https://github.com/TykTechnologies/tyk-operator/blob/master/.github/workflows/cut_release.yml

We have a new GitHub action - called cut_release.

It is triggered manually from GitHub Actions.

Screenshot 2020-12-02 at 13 41 04

We specify the version of the release we want to cut - and apply that to the master branch.

When we manually trigger the release, the action should:

- [ ] Change the tagged version in the helm chart.yaml https://github.com/TykTechnologies/tyk-operator/blob/master/helm/Chart.yaml#L5
- [ ] Tag the master branch & push the tags.

This will trigger a goreleaser github action / build pipeline to push to Dockerhub - once master branch is pushed.

---

The current flow is a little manual & as such, if i forget to bump the Chart.yaml or something - then my release is broken.

alephnull:
Fixes #191.

Largely untested, at least in terms of pushing a version.


asoorm:
@alephnull 

But you removed the manual action and input?

```
on:
  workflow_dispatch:
    inputs:
      tag:
        description: 'Release tag'
        required: true
```

And the operator should use that tag to bump the helm version and push the tag to GitHub.

alephnull:
This sweet ass sed line should do the business
- famous last words

alephnull:
@asoorm We'll have to push some tags to test. How about I push some tags suffixed `-test`?

asoorm:
@alephnull bit concerned that we are using Semver - and pushing `test` prefixed / suffixed will not work.

```
https://github.com/TykTechnologies/tyk-operator/blob/master/.goreleaser.yml#L18-L22
```

Can we release it as `v0.3.1`? And do we have ability to delete the Dockerhub image if it goes wrong?

PR #199, opened 2020-12-04
pin the operator image tag

sedkis:


asoorm:
Thanks @sedkis - I'm personally not a fan of doing things / making decisions auto-magically on-behalf of the user. Default `latest` is expected functionality. Users expect to pin if they want / need to. This is standard practice.

Unless we get a compelling reason to change status quo - I think we can leave it as is for time being.

If we did pin, then we would need to build this logic into the CI to auto update the values.yaml.

Issue #134, opened 2020-10-29
feat: defaulting webhooks for security policies


asoorm:
https://github.com/TykTechnologies/tyk-operator/blob/master/config/samples/httpbin_protected_policy.yaml#L15-L33

Disable Rate Limits, Quotas, Throttle by default.
Ensure that any configurations are explicit.

We can implement this from now, but ultimately - we need the following fix: https://github.com/TykTechnologies/tyk-operator/issues/114 for policies to be useful.

asoorm:
Closed by https://github.com/TykTechnologies/tyk-operator/pull/200

PR #200, opened 2020-12-08
Idempotent security policy

### LINKED ISSUE #182: bug: security policies are not idempotent
Apply an API
Apply a security policy
Apply the security policy again and you will see that the CR is modified.

image

```
kubectl apply -f ./config/samples/httpbin_protected.yaml
kubectl apply -f ./config/samples/httpbin_protected_policy.yaml
kubectl apply -f ./config/samples/httpbin_protected_policy.yaml
```

The second time we apply the policy CR, kubectl should tell us that the policy was unchanged, rather than configured.

This implies that the reconciler is modifying the security policy custom resource.

This may be resolvable by introducing defaulting webhook for security policies: https://github.com/TykTechnologies/tyk-operator/issues/134

gernest:
Fixes #182


helm/crds/crds.yaml
```diff
@@ -1602,7 +1602,6 @@ spec:
               description: Number of retries before returning error.   omit or "-1" for unlimited
               type: integer
           required:
-          - access_rights_array
```
asoorm, line 1605: Why did we remove `access_rights_array`? Surely the old method is deprecated?

asoorm:
Fixes #134

PR #201, opened 2020-12-08
Ingress Support

asoorm:


docs/ingress.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Sample Template ApiDefinition resource
+
+```yaml
+apiVersion: tyk.tyk.io/v1alpha1
+kind: ApiDefinition
+metadata:
+  name: myapideftemplate
+  labels:
+    - isIngressTemplate: true
```
gernest, line 33: I am having a second thought about this. Seeing that in our ingress we have annotation

```
tyk.io/template: myapideftemplate
```

Maybe we should have something  similar like 

```
  labels:
    - tyk.io/template: true
```

I have seen this pattern even on the k8s  api's (can't remember where) but keeping it consistent is a good think so we don't have to introduce too many labels. We will only have `tyk.io/template` , which acts as a filter on `APIDefinition` and used to annotate the ingress resource.

gernest: Or go even further and use `tyk.io/template: myapideftemplate` for both ingress and template API def, and we matches them to link them together.

asoorm:
```
kind: ApiDefinition
```

But we are already inside a Tyk object. So we don't need to be careful about prefixing with `tyk.io`. Usually you add the prefix to resources that you don't own in order to play nice with other tooling.

e.g. on the Ingress resource: `cert-manager.io/issuer: my-issuer` because Ingress is not owned by cert-manager.

asoorm: If other tools want to extend / integrate with a Tyk ApiDefinition resource - they are welcome to prefix their annotations.

gernest: That makes sense.

@asoorm what about 

```
  labels:
    - template: true
```

gernest: It's not like we will be having different kinds of `APIDefinition` templates, this will only mark the Object as a template.

controllers/ingress_controller.go
```diff
@@ -80,102 +116,142 @@ func (r *IngressReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
 
-	ingressClass := &v1beta1.IngressClass{}
-	if err := r.Get(ctx, types.NamespacedName{Name: *desired.Spec.IngressClassName, Namespace: req.Namespace}, ingressClass); err != nil {
-		println(err.Error())
-		return ctrl.Result{}, err
-	}
+	for _, rule := range desired.Spec.Rules {
+		hostName := rule.Host
+		for _, p := range rule.HTTP.Paths {
+			apiName := strings.TrimRight(strings.ReplaceAll(fmt.Sprintf("%s-%s-%s-%s", req.Name, req.Namespace, hostName, p.Path), "/", ""), "-")
```
asoorm, line 141: Anybody got any ideas for a decent API Naming convention?

gernest: can we move this to a separate function first? If I understand correctly we will be deriving the names here  for both creating and updates so this function should be able to generate unique names for each Ingress rule.

controllers/ingress_controller.go
```diff
@@ -80,102 +116,142 @@ func (r *IngressReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+					Name:      apiName,
+					Namespace: req.Namespace,
+					Labels:    lbls,
+				},
+				Spec: template.Spec,
+			}
+
+			api.Spec.Name = apiName
+			api.Spec.Proxy.ListenPath = p.Path
+			api.Spec.Proxy.TargetURL = fmt.Sprintf("http://%s.%s.svc.cluster.local:%d", p.Backend.ServiceName, namespacedName.Namespace, p.Backend.ServicePort.IntValue())
```
gernest, line 153: What does this do? In tyk perspective, Target actually means upstream. This is user provided, we shouldn't be doing anything on this field. This will be something like `http://httpbin.org` now if they want to target cluster services they should explicitly state it and dns will resolve accordingly.

docs/ingress.md
```diff
@@ -0,0 +1,77 @@
...
+ rules:
+  - http:
+     paths:
+      - path: /httpbin
+        pathType: Prefix
+        backend:
+         service:
+          name: httpbin
+          port:
+           number: 8000
```
gernest, line 71: what does this do?  Is this a tyk api  with that name/port. Because that will be in conflict with the whole template concept, however we can leverage this to allow customers to provide a unique `service` name which will be the `APIDefinition` name, this will remove the need to use complex schemes to generate api names.

```diff
diff --git a/controllers/ingress_controller.go b/controllers/ingress_controller.go
index b14a274..70f79f6 100644
--- a/controllers/ingress_controller.go
+++ b/controllers/ingress_controller.go
@@ -148,7 +148,7 @@ func (r *IngressReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 				Spec: template.Spec,
 			}
 
-			api.Spec.Name = apiName
+			api.Spec.Name = p.Backend.ServiceName
 			api.Spec.Proxy.ListenPath = p.Path
 			api.Spec.Proxy.TargetURL = fmt.Sprintf("http://%s.%s.svc.cluster.local:%d", p.Backend.ServiceName, namespacedName.Namespace, p.Backend.ServicePort.IntValue())
 
```

gernest (changes requested):

docs/ingress.md
asoorm: The idea is that the template can be used/shared by many ingress resources and ingress rules.

controllers/ingress_controller.go
asoorm: The whole purpose of ingress is to target cluster services

docs/ingress.md
gernest: Isn't this what is happening? `Spec: template.Spec,`? we creating concrete resources here based on the template. And the name is replaced with actual name the user specified in the ingress backend.

controllers/ingress_controller.go
gernest: >The whole purpose of ingress is to target cluster services?

Yes, and in our case they are `api's` behind  tyk in the cluster.

we are `Incomming traffic` `=>` `tyk/api-service-endpoint`

gernest: we are exposing the api's to the outside world. This is what I think.

asoorm: @gernest after catchup - is it ok to resolve this conversation?

controllers/ingress_controller.go
```diff
@@ -221,6 +221,10 @@ func (r *IngressReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	return ctrl.Result{}, nil
 }
 
+func (r *IngressReconciler) buildAPIName(nameSpace, name, hostName, path string) string {
+	return strings.TrimRight(strings.ReplaceAll(fmt.Sprintf("%s-%s-%s-%s", nameSpace, name, hostName, path), "/", ""), "-")
```
gernest, line 225: Looking at this I am thinking we can play it a bit safer. Because path docs says

> 	// Path is matched against the path of an incoming request. Currently it can
	// contain characters disallowed from the conventional "path" part of a URL
	// as defined by RFC 3986. Paths must begin with a '/'. When unspecified,
	// all paths from incoming requests are matched.
	// +optional

A simple replacement of `/` with `""` won't do and much and will complicate things

```suggestion
	return fmt.Sprintf("%s-%s-%s", nameSpace, name, encodeNS(hostName+path))
```

This will be just `{namespace}-{name}-{base64 encoded combo of host+path}`

we can revisit this any other time if requirements changes.

gernest (changes requested):

controllers/ingress_controller.go
asoorm:
```
a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')"
```

For the purposes of this PR - I will leave as-is, but we will need to fix it further down the line for sure. e.g. we don't currently support regexp in the listen path for the operator and im not sure we will need to.

gernest: > we don't currently support regexp in the listen path for the operator and im not sure we will need to
Do we have a validation for this? And `listen_path` is bread and butter of the gateway apidef. This is a very bold statement.

asoorm: Ingress doesn't allow regexp in the listen path. So we wouldn't need to do it for ApiDefinition either.

PR #228, opened 2020-12-31
block template deletion when there are ingresses depending on it.

### LINKED ISSUE #226: delete template api definition should block if ingress resources depend upon it
At present - ingress resources can depend on ingress templates (api definitions with template annotation).

If I try to delete the ingress template, and there are referenced ingress dependencies, this means that a reconcile on the ingress resource will cause ingress to break, as they will have no template to be applied against.

### Recreation Steps

```
k get tykapis
NAME                                 DOMAIN     LISTENPATH   PROXY.TARGETURL                                 ENABLED
default-httpbin-ingress1-e0bf94ae1   httpbin1   /            http://httpbin.default.svc.cluster.local:8000   true
default-httpbin-ingress2-3e8322b04   httpbin2   /            http://httpbin.default.svc.cluster.local:8000   true
myapideftemplate                                             http://example.com                              true
```

```
k delete tykapis myapideftemplate
apidefinition.tyk.tyk.io "myapideftemplate" deleted
```

```
k get tykapis
NAME                                 DOMAIN     LISTENPATH   PROXY.TARGETURL                                 ENABLED
default-httpbin-ingress1-e0bf94ae1   httpbin1   /            http://httpbin.default.svc.cluster.local:8000   true
default-httpbin-ingress2-3e8322b04   httpbin2   /            http://httpbin.default.svc.cluster.local:8000   true
```

```
k get ingress httpbin-ingress1 -o json | jq '.metadata.annotations'
{
  "kubectl.kubernetes.io/last-applied-configuration": "TRUNCATED - NOT IMPORTANT",
  "kubernetes.io/ingress.class": "tyk",
  "tyk.io/template": "myapideftemplate" <-----------------
}
```

### Expected functionality

```
Scenario: Delete an Ingress Template
Given: I have an ingress template `foo`
  And: I have an ingress resource referencing `annotations/tyk.io/template: foo`
When: I delete ingress template `foo`
Then: The apiDefinition controller should log the problem, return an error, allowing K8s to retry by requeueing
```

### Proposed implementation

In order to solve this, maybe we need a finalizer in the api definition template resource.

gernest:
Closes #226

### DIFF
--- a/controllers/apidefinition_controller.go
+++ b/controllers/apidefinition_controller.go
@@ -21,6 +21,7 @@ import (
 	"time"
 
 	"github.com/TykTechnologies/tyk-operator/pkg/cert"
+	"github.com/TykTechnologies/tyk-operator/pkg/keys"
 
 	tykv1alpha1 "github.com/TykTechnologies/tyk-operator/api/v1alpha1"
 	"github.com/TykTechnologies/tyk-operator/pkg/universal_client"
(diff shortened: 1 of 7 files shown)


asoorm (approved):
Looks good - thanks

PR #234, opened 2021-01-04
deterministic api control

### LINKED ISSUE #71: track: deleting API leaves artifacts inside the organization document
https://tyktech.atlassian.net/browse/TT-855

Creating an API using the tyk-operator creates 2 artifacts inside the organization object.

The Original API ID, and the Custom API ID.

Screenshot 2020-10-07 at 14 34 57

When deleting the API by the custom ID, the original ID remains as artifact.

Screenshot 2020-10-07 at 14 37 48

---

I think that the problem is not necessarily with Deleting an API, but the act of Updating the API (to provide a custom ID) fails to update the organization object with the new reference and just creates a new entry.

### LINKED ISSUE #231: api ids should be deterministic
3 use cases which Tyk Operator was designed for:

1. Migrate from Tyk CE to Tyk Pro
2. Backup & Restore (disaster recovery)
3. Ability to upgrade operator versions

All above should work without keys stopping working.

---

The API ID was designed to be deterministic `base64(namespace/name)`.

1. create protected API
2. create key granting access to that API
3. delete API
4. key stops working
5. re-create the API
6. key should continue working again.

---

The Policy was also designed to be deterministic. 

Currently works, but must continue working

1. create protected API
2. create a policy granting access to that API
3. create key granting access to that policy
4. delete policy
5. key stops working
6. re-create the policy
6. key should continue working again.

---

When customers upgrade from 0.1 to 0.2, at the moment, their keys will all stop working because apiid will change from b64encoded namespace/name to mongo ids.

gernest:
rely on deterministic `api_id` scheme for both community and pro edition

fixes #231 
fixes #71

### DIFF
--- a/bdd/godogs_test.go
+++ b/bdd/godogs_test.go
@@ -24,7 +24,7 @@ import (
 const (
 	namespace      = "bdd"
 	k8sTimeout     = time.Second * 10
-	reconcileDelay = time.Second * 10
+	reconcileDelay = time.Second * 5
 )
 
 var gwNS = fmt.Sprintf("tyk%s-control-plane", os.Getenv("TYK_MODE"))
--- a/bdd/godogs_test.go
+++ b/bdd/godogs_test.go
@@ -254,7 +254,7 @@ func wait(ts time.Duration) func(err error) error {
 		}
 		time.Sleep(ts)
 		cmd := exec.CommandContext(context.Background(), "kubectl", "get", "tykapis", "-n", namespace)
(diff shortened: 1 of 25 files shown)


gernest:
@asoorm can you please help test if this PR actually address #71  too?

asoorm:
> @asoorm can you please help test if this PR actually address #71 too?

NICE!!!

asoorm (approved):

PR #252, opened 2021-02-01
handle float64 field

gernest:
use `json.Number` for threshold_percent which is a float64


sedkis:
Tested, now working:

asoorm (approved):

asoorm (changes requested):
Need to update the documentation showing that Circuit Breaker now works

https://github.com/TykTechnologies/tyk-operator/blob/master/docs/api_definitions.md#apidefinition---endpoint-middleware

Screenshot 2021-02-03 at 10 12 06

Even better to have a BDD example also.

api/v1alpha1/float64.go
```diff
@@ -0,0 +1,81 @@
+package v1alpha1
+
+import (
+	"encoding/json"
+	"fmt"
+	"sync"
+)
+
+//Float64 is a work around to allow representing floating points as strings
+// +kubebuilder:validation:Pattern="^(?:[-+]?(?:[0-9]+))?(?:\\.[0-9]*)?(?:[eE][\\+\\-]?(?:[0-9]+))?$"
```
asoorm, line 10: Am i right? This kubebuilder validation pattern doesn't seem to match the pattern generated in the CRDs or the tests?

```
pattern: ^0\.\d+|1\.0$ 
```

gernest: This is for a full `float64` value, `pattern: ^0\.\d+|1\.0$ ` is just for matching `Percent` . A `Percent` is still a valid `float64`

api/v1alpha1/float64.go
```diff
@@ -0,0 +1,81 @@
...
+// Percent describes a percentage value expressed as a float. This is a positive
+// decimal value that is less than 1
+//+kubebuilder:validation:Pattern="^0\\.\\d+|1\\.0$"
+type Percent string
+
+// MarshalJSON returns a json string for p. This is a string for normal
+// operations and a float64 when marshalling for dashboard or gateway
+func (p Percent) MarshalJSON() ([]byte, error) {
+	return Float64(p).MarshalJSON()
+}
```
gernest, line 71: @asoorm check here we are  using `Float64. MarshalJSON` . So it is okay for some values that require actual `float64` to use the `Float64` type.

api/v1alpha1/float64.go
asoorm: Ok - Can we have some tests for this scary looking regexp? or if we copied and pasted it from somewhere, what do you think if we provide link to the src so we can track it?

gernest: done c680f77

Added link to the source https://github.com/asaskevich/govalidator/blob/7a23bdc65eef5f3783e782b436f3065eae3fc72d/patterns.go#L19

Tests are included there, it is a go validation library.

asoorm: awesome

PR #253, opened 2021-02-09
Ensure dev make rules works for kind and minikube

### LINKED ISSUE #203: makefile is currently tightly coupled to local kind cluster development
We now have helpers to improve developer workflow such as:

`make boot-pro IMG=tykio/tyk-operator:test`.

This is great, however it makes it difficult to use against a minikube cluster for local development, or a real K8s cluster in the cloud when testing Ingress for example.

We need to remove the hard dependency on `kind load docker-image` and at least support minikube for local development too.

```
.PHONY: install-operator-helm
install-operator-helm: cross-build-image manifests helm
	@echo "===> installing operator with helmr"
	kind load docker-image ${IMG} --name=${CLUSTER_NAME}
	helm install ci ./helm --values ./ci/helm_values.yaml -n tyk-operator-system --wait
```

gernest:
Fixes #203

For case of minikube we use `cache` command to update the development image into minikube


asoorm:
Loving your Go Scripting!

asoorm (approved):

PR #259, opened 2021-02-18
add prerequisite for pro deployments

gernest:
```
    "policies": {
        "allow_explicit_policy_id": true
    },
```

Is required to be set in `tyk.conf` for SecurityPolicy resources to properly
work.

This is because the operator uses `id` to refer to security policy id's while
the gateways will by default use `_id` to refer to policy id's and when `policies.allow_explicit_policy_id` is true the gateway will use `id` when
it is provided to identify policies.

### DIFF
--- a/docs/installation/installation.md
+++ b/docs/installation/installation.md
@@ -16,6 +16,12 @@ Before running the operator
 * A secret in the namespace of your operator deployment telling the operator how to communicate with Tyk
 * The CRDs must be registered with the Kubernetes apiserver
 * cert-manager must be installed
+* If you are using `pro` edition make sure in your gateway's `tyk.conf` `policies.allow_explicit_policy_id` is set to true
+```json
+    "policies": {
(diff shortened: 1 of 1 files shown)


docs/installation/installation.md
```diff
@@ -16,6 +16,12 @@ Before running the operator
 * A secret in the namespace of your operator deployment telling the operator how to communicate with Tyk
 * The CRDs must be registered with the Kubernetes apiserver
 * cert-manager must be installed
+* If you are using `pro` edition make sure in your gateway's `tyk.conf` `policies.allow_explicit_policy_id` is set to true
+```json
+    "policies": {
+        "allow_explicit_policy_id": true
+    },
+```
```
asoorm, line 24: This is great. I feel like in k8s environment, an env var would more likely be used.

gernest: We don't deploy tyk so it's up to our customers to decide how the set the value.

PR #261, opened 2021-02-27
Ourbound port exception for dash comms on istio

joshblakeley:
When we use the operator in an injected namespace on Istio we will need to allow non intercepted requests to the dashboard (on port 3000 usually).

This annotation lets that happen when in that environment but does not change anything otherwise.

See related tyk-helm-chart PR that also brings this work together: https://github.com/TykTechnologies/tyk-helm-chart/pull/80


gernest:
This has been properly addressed  in #262

PR #269, opened 2021-03-12
Tyk helm chart

### LINKED ISSUE #268: Use official tyk helm charts for CI
We use our own deployment of tyk  for both ci and development, however our customers uses official charts for tyk deployment and they are facing issues that don't suffice when developing this project.

This will allow us to improve the tyk charts  and streamline experimenting with the operator with  standard tyk deployments

gernest:
closes #268 

Waiting for 
- [x] https://github.com/TykTechnologies/tyk-helm-chart/pull/93
- [x] https://github.com/TykTechnologies/tyk-helm-chart/issues/94
- [x] https://github.com/TykTechnologies/tyk-helm-chart/issues/92
- [ ] ~https://github.com/TykTechnologies/tyk-helm-chart/pull/72~

## NOTE

This PR changes `TYK_MODE` env var value for Tyk community edition to `ce` instead of `oss`.


pkg/environmet/env.go
```diff
@@ -96,7 +96,7 @@ func (e *Env) Parse() error {
 		return fmt.Errorf("environment vars %v are missing", ls)
 	}
 	switch e.Mode {
-	case "oss", "pro":
+	case "ce", "pro":
```
asoorm, line 99: Are you sure this should be `ce` rather than `oss`?

asoorm:
Just a thought / consideration:

The Gateway (proxy) service needs to publish 2 ports - 80 and 443 for Ingress.

gernest:
> The Gateway (proxy) service needs to publish 2 ports - 80 and 443 for Ingress.

We don't have  tests for ingress yet, so I don't think this is necessary for now.

pkg/environmet/env.go
gernest: I have refactored, we had two env vars that were refering to the same thing. Some places used `ce` some places used `oss`. I decided to consolidate and use `ce` everywhere.

asoorm: does that introduce a breaking change though? can we change it so that we can accept `ce` or `oss`? we also need a docs update:

https://github.com/TykTechnologies/tyk-operator/blob/master/docs/installation/installation.md#tyk-operator-configuration

gernest: @asoorm 

The docs have been updated here  cf74582
Added backward compatibility here 9b31fba so this is no longer a breaking change

PR #271, opened 2021-03-23
support oauth2 client credentials for api protection

asoorm:


api/v1alpha1/apidefinition_types.go
```diff
@@ -699,12 +720,26 @@ type BasicAuthMeta struct {
...
+
+	// AllowedAuthorizeTypes is an array of allowable `response_type` parameters `token` or authorization code `code`.
+	// Choose token for client_credentials or implicit grant types.
+	AllowedAuthorizeTypes []AuthorizeTypeEnum `json:"allowed_authorize_types"` // osin.AuthorizeRequestType
+
+	// Login form to handle user login.
+	AuthLoginRedirect string `json:"auth_login_redirect,omitempty"`
 }
 
+// +kubebuilder:validation:Enum=authorization_code;refresh_token;password;client_credentials
```
gernest, line 737: according to osin, `__implicit` and `assertion` are missing

```suggestion
// +kubebuilder:validation:Enum=authorization_code;refresh_token;password;client_credentials;assertion;__implicit
```

gernest (changes requested):
Looks good. I just left a comment.

api/v1alpha1/apidefinition_types.go
asoorm: Screenshot 2021-03-29 at 10 34 20

Thanks for the suggestion. According to Tyk Docs - we don't support implicit grant. And i don't know what `assertion` is. I also am not so keen on users relying on undocumented features. So would prefer not to include for the time being.

gernest (approved):
LGMT

PR #276, opened 2021-04-02
reducing permissions of secret controller

asoorm:
https://github.com/TykTechnologies/tyk-operator/discussions/275


gernest (approved):
LGMT

Issue #251, opened 2021-02-01
build and publish latest docker tag upon merge to master


asoorm:
At present, latest docker image is only built when we cut a release.
This causes problems, because master branch of tyk-operator (CRDs) 
might then be out-of-sync with the tyk-operator:latest docker tag. We
also must consider the use-case for making release-candidates available.

Currently we have goreleaser.yaml with

```
image_templates:
  "tykio/tyk-operator:latest"
  "tykio/tyk-operator:v{{ .Major }}"
  "tykio/tyk-operator:v{{ .Major }}.{{ .Minor }}"
  "tykio/tyk-operator:{{ .Tag }}"
```

Which we is executed when we push a release tag like `v0.5.0`

What we need is:

When we marge/push any change to master, a workflow should be triggered that only builds latest docker image

```
image_templates:
  "tykio/tyk-operator:latest"
```

And when we push a release candidate `v0.5.0-rc1` only build docker image for that specific tag

```
image_templates:
  "tykio/tyk-operator:{{ .Tag }}"
```

When we push a release tag `v0.5.0` don't tag latest

```
image_templates:
  "tykio/tyk-operator:v{{ .Major }}"
  "tykio/tyk-operator:v{{ .Major }}.{{ .Minor }}"
  "tykio/tyk-operator:{{ .Tag }}"
```

alephnull:
https://github.com/TykTechnologies/tyk-operator/blob/master/.github/workflows/release_tag.yaml#L9-L11 is setup such that it is triggered when a tag is pushed. It can be configured to [operate on master](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestbranchestags).

[goreleaser](https://github.com/TykTechnologies/tyk-operator/blob/master/.goreleaser.yml#L19) has been configured to always push the latest tag on _every_ image it pushes.

alephnull:
Please add this to the DevOps jira if you'd like one of us to work on it, but it is a small change that can be made by anyone really. Happy to hop on a zoom with anyone who is interested.

gernest:
@alephnull I created a jira ticket to track this with additional details https://tyktech.atlassian.net/browse/TT-1617 
please take a look at it.

PR #292, opened 2021-04-21
Build nightly release images/artifacts

gernest:
This PR adds support for releasing nightly builds

Related #251

Now we have two kinds of releases

# Normal releases

These are triggered  with a semver tag e.g `v0.6.0` They are handled by goreleaser configuration
`.goreleaser.yml`

# Nightly releases

These are triggered with semver tag with nightly prerelease  like `v0.6.0-nightly.b7949b1` and handled
by goreleaser configuration `.goreleaser_nightly.yml`.

The purpose of this rellease is to have artifacts that reflects latest master commit on the master branch.This ensures the docker image `tykio/tyk-operator:latest` will always be pointing to the latest master branch.

When a new commit is pushed to master that changes `.go` files and it doesn't match the latest tag's commit, a new tag is created of the format `{Tag}-nightly.{short_commit_hash}` that will trigger the nightly build.

# Docker image

The nightly build updates `tyk-operator:latest` docker image to always point to the latest master commit.

# How it was tested

I manually created a repo https://github.com/gernest/create-tag-workflow to test the script that generates the nightly tag from the master commits and verified it works.

The release workflow is the same(unchanged) so everything else should work as expected.


asoorm:
I'm not so sure how this solves the problem. 
We need to build and push `tyk-operator:latest` whenever we merge a PR to master branch.

The challenge that we have, is that users read the docs from the **master** branch.
They may have `latest` installed, but latest and master don't match.

gernest:
simply replacing `tykio/tyk-operator:nightly` with `tykio/tyk-operator:latest` should do the trick. 

This is still a draft, I'm collecting my thoughts.

asoorm (approved):
LGTM

PR #308, opened 2021-05-09
chore: use context.Context in universal client

gernest:
This PR updates  interfaces exposed by Universal client to use `context.Context`

## Why ?
Context awareness helps in

- properly manage http.Client (Timeouts/deadlines and cancelation)
- resource cleanup (in case of offloading expensive ops in goroutines)
- raise awareness of execution environment . This is important since we are now thinking about multi tenancy.

The interfaces updated are `UniversalApi` `UniversalCertificate` and `UniversalSecurityPolicy`


asoorm (approved):
LGTM

PR #309, opened 2021-05-09
add custom_middleware_bundle on APIDefinitionSpec

### LINKED ISSUE #306: Support the custom_middleware_bundle field
## Expected Behavior

As an API owner i should be able to target bundles being served from my plugin server

gernest:
Other plugin related fields are enabled/present so there is no need for
this not to be present as well.

Closes #306

### DIFF
--- a/api/v1alpha1/apidefinition_types.go
+++ b/api/v1alpha1/apidefinition_types.go
@@ -627,8 +627,8 @@ type APIDefinitionSpec struct {
 	//DisableRateLimit       bool                `json:"disable_rate_limit"`
 	//DisableQuota           bool                `json:"disable_quota"`
 
-	CustomMiddleware MiddlewareSection `json:"custom_middleware,omitempty"`
-	//CustomMiddlewareBundle string              `json:"custom_middleware_bundle"`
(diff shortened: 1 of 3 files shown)


api/v1alpha1/apidefinition_types.go
```diff
@@ -627,8 +627,8 @@ type APIDefinitionSpec struct {
 	//DisableRateLimit       bool                `json:"disable_rate_limit"`
 	//DisableQuota           bool                `json:"disable_quota"`
 
-	CustomMiddleware MiddlewareSection `json:"custom_middleware,omitempty"`
-	//CustomMiddlewareBundle string              `json:"custom_middleware_bundle"`
+	CustomMiddleware       MiddlewareSection `json:"custom_middleware,omitempty"`
+	CustomMiddlewareBundle string            `json:"custom_middleware_bundle"`
```
asoorm, line 631: Would be good to have some documentation for these fields - as we are here. What does `CustomMiddleware` and `CustomMiddlewareBundle` do? How do I use it? is there an example? 

Also, why not `omitempty`? ``json:"custom_middleware_bundle,omitempty"``

asoorm (changes requested):

asoorm (approved):

PR #312, opened 2021-05-17
chore: remove lots of dead code

gernest:
There is a lot of code that is un documented  that does nothing and some which we don't support yet.

## Why ?

- We haven't fully upgraded to using `kubebuilder v3`  these resources were adding lots of noice and make things harder to upgrade since if they are around I am forced to support them too.
- They look scary and can discourage people to contribute
- We will bring them back whenever we feel like it is the right thing to do, however lots of thoughts needs to be done to get things right
- For the case of `Gateway` it is beyond the scope of the Operator, we have `tyk-helm-charts` repo for this case.

## What is being removed

- Organization
- Webhook
- Gateway

**NOTE**
This is blocking me on #285  since I need to fully upgrade to `v3` to leverage new api for configuring operators.


asoorm:
Screenshot 2021-05-17 at 08 53 39

Need to remove reference to webhook from README.md

asoorm (approved):
LGTM

PR #316, opened 2021-05-18
proper upgrade to kubebuilder v3

gernest:
| tool | version|
|-------|---------|
| operator-sdk| v1.7.2 |
|controller-gen | v0.4.1|
| kustomize |v3.8.7|
| helm|v3.5.4|


helm/values.yaml
```diff
@@ -1,51 +1,25 @@
...
-  #   cpu: 100m
-  #   memory: 128Mi
-
-nodeSelector: {}
-
-tolerations: []
-
-affinity: {}
-
-clusterDnsDomain: cluster.local
```
asoorm, line 51: Why are we removing these from the values.yaml?

gernest: - `podSecurityContext` is not removed
`resources` is though, so thanks for catching this.

The rest were never used so just redundant.

asoorm (approved):
LGTM

PR #317, opened 2021-05-20
per reconcile loop http api context

gernest:
This PR refactors universal client to use `context.Context` to target specific environment. This context is sent per reconciliation loop, to easy transition to multi tenancy.

This refactoring is needed for multi tenancy work, for now we pass the default environment .

There is no breaking change in this PR.


asoorm (approved):
lgtm

PR #329, opened 2021-06-17
add support for portal catalogue

### LINKED ISSUE #266: Support for updating the API Catalog
## Expected Behavior
As an API developer I would like to have the ability to have the OpenAPI documentation managed by the Tyk Operator

## Current Behavior
Currently the Tyk Operator does not support updating the API Catalog.
The only way to update the Swagger documentation is by manually removing and then adding the documentation in the Tyk Dashboard.

## Possible Solution
Add the possibility to link in the API Definition to a config map containing the Swagger specification.

gernest:
Closes #266

Successor to and closes #290


api/model/portal_catalogue.go
```diff
@@ -7,53 +7,116 @@ type APICatalogue struct {
...
+	Name string `json:"name,omitempty"`
+
+	// AuthType displays as a badge next to the name of the API
+	AuthType AuthType `json:"auth_type,omitempty"`
+
+	// Show toggles visibility of the API in the portal catalogue
+	Show bool `json:"show,omitempty"`
+
+	// TODO: I don't think this is exposed to the default portal templates.
+	ShortDescription string `json:"short_description,omitempty"`
```
asoorm, line 26: @buger do you know anything about this field? It seems that the short_description has no effect on visibility in the templates. It seems to have no effect even when configured in the GUI.

api/model/portal_catalogue.go
```diff
@@ -7,53 +7,116 @@ type APICatalogue struct {
...
+	DisableLogin         bool         `json:"disable_login,omitempty"`
+	DisableSignup        bool         `json:"disable_signup,omitempty"`
+	DisableAutoLogin     bool         `json:"disable_auto_login,omitempty"`
+	CatalogueLoginOnly   bool         `json:"catalogue_login_only,omitempty"`
+	OAuthUsageLimit      int          `json:"oauth_usage_limit,omitempty"`
+	Email                string       `json:"email,omitempty"`
+	MailOptions          *MailOptions `json:"mail_options,omitempty"`
+	DCROptions           *DCROptions  `json:"dcr_options,omitempty"`
+	EnableDCR            bool         `json:"enable_dcr,omitempty"`
+	Override             bool         `json:"override,omitempty"`
```
asoorm, line 77: These fields need to be documented

api/model/portal_catalogue.go
buger: Most likely not. Try creating API catalog via UI, filling all fields - if this field not used, ignore it.

api/v1alpha1/portalapicatalogue_types.go
```diff
@@ -0,0 +1,67 @@
...
+import (
+	"github.com/TykTechnologies/tyk-operator/api/model"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// PortalAPICatalogueSpec defines the desired state of PortalAPICatalogue
+type PortalAPICatalogueSpec struct {
+	OrgID              string                        `json:"org_id,omitempty"`
+	Email              string                        `json:"email,omitempty"`
+	APIDescriptionList []*PortalCatalogueDescription `json:"apis,omitempty"`
```
asoorm, line 28: Needs docs

asoorm (commented):
Awesome work @gernest . I think we just need some of the most popular fields commented to document the api. And disable fields that we don't know what they do yet.

api/model/api.go
```diff
@@ -261,6 +262,12 @@ type Target struct {
 	Namespace string `json:"namespace"`
 }
 
+func (t *Target) Parse(v string) {
+	p := strings.Split(v, "/")
+	t.Namespace = p[0]
+	t.Name = p[1]
```
asoorm, line 268: This could be made more safe. e.g. what happens if the string doesn't contain a `/`?

asoorm (approved):
🥳

PR #337, opened 2021-07-23
docs: add workaround for multi tenancy without  default environment

gernest:


docs/operator_context.md
```diff
@@ -106,4 +106,36 @@ spec:
...
+
+With  our helm charts we have
+```
+envFrom:
+  - secretRef:
+      name: tyk-operator-conf
+```
+That loads the default env from secret `tyk-operator-conf` . This might not be desired if you are already using operator context for resources.
+
+One work around is to set dummy values that way the operator will start, and provide valid context when appllying resources that will be used by the operator for reconciliation.
```
asoorm, line 123: typo: `appllying` -> `applying`

asoorm (approved):

Issue #347, opened 2021-08-13
deleting an OperatorContext error/requeue if it is referenced by other resources


asoorm:
Create 2 resources:

OperatorContext
ApiDefinition - which references the context

Delete the OperatorContext - This should throw an error and retry on exponential backoff, because it is depended upon by the ApiDefinition Object.

## Recreation Steps:

1. Create OperatorContext
```
cat <<EOF | kubectl apply -f -
apiVersion: tyk.tyk.io/v1alpha1
kind: OperatorContext
metadata:
  name: mycontext2
spec:
  env:
    auth: 7937405e587d491d7a5025dce725e838
    mode: pro
    org: 6116bed8d3626e78fa9e1671
    url: http://2d113e1dbc09.ngrok.io
    ingress:
      httpPort: 8080
      httpsPort: 8443
EOF
```

2. Create ApiDefinition
```
cat <<EOF | kubectl apply -f -
apiVersion: tyk.tyk.io/v1alpha1
kind: ApiDefinition
metadata:
  name: httpbin2
spec:
  contextRef:
    name: mycontext2
  name: HttpBin Service
  protocol: http
  use_standard_auth: true
  proxy:
    listen_path: /httpbin2
    strip_listen_path: true
    target_url: http://httpbin.org
  active: true
  version_data:
    default_version: Default
    not_versioned: true
    versions:
      Default:
        name: Default
EOF
```

3. Delete OperatorContext
```
kubectl delete operatorcontexts.tyk.tyk.io mycontext2
operatorcontext.tyk.tyk.io "mycontext2" deleted               <------------ THIS SHOULD FAIL
```

4. We are now unable to operate on ApiDefinition which depends on the OperatorContext which was deleted.

PR #375, opened 2021-11-25
[TT-3673] Fix operator context issue

komalsukhani:
### Changed Operator Context Status
 Updated **Status**  to store list of all the resources referring to it. It will contain following fields, which are of types **[]model.Target**
      
- linked_api_definitions     
- linked_api_descriptions    
- linked_portal_catalogues 
- linked_security_policies    
- linked_portal_configs 

### Added Controller for Operator Context
Before deleting an operator context it will check if any other resource refers to it. If it does, then delete operation is re queued.

### Fixed linter issues

- Removed unused code
- Added proper spacing
- Removed deprecated function call

### Updated kind config
It will create 2 worker nodes too

## Related Issue

[#347](https://github.com/TykTechnologies/tyk-operator/issues/347)

## Motivation and Context

Deleting an operator context, while it is being referred by other resources, leaves them in broken state.

Added tests for operator context create and delete scenarios.

## Types of changes

## Checklist

- [x] Modules and vendor dependencies have been updated; run `go mod tidy && go mod vendor` in applicable directories.
  - [ ] `gofmt -s -w .`
  - [ ] `go vet ./...`
  - [ ] `golangci-lint run`


.vscode/launch.json
```diff
@@ -0,0 +1,24 @@
...
+            "name": "Debug Package",
+            "type": "go",
+            "request": "launch",
+            "mode": "debug",
+            "env":{
+                "TYK_MODE":"ce"
+            }
+        }
+    ]
+}
```
asoorm, line 24: What's this file? looks like .vscode directoryshould be added to .gitignore?

komalsukhani: Opps sorry. I will remove it.

api/v1alpha1/operatorcontext_types.go
```diff
@@ -110,3 +117,53 @@ type OperatorContextList struct {
...
+
+func (opStatus *OperatorContextStatus) RemoveLinkedAPIDefinition(target model.Target, log logr.Logger) {
+	for i, apiDef := range opStatus.LinkedApiDefinitions {
+		if apiDef.Namespace == target.Namespace && apiDef.Name == target.Name {
+			opStatus.LinkedApiDefinitions = append(opStatus.LinkedApiDefinitions[:i], opStatus.LinkedApiDefinitions[i+1:]...)
+
+			return
+		}
+	}
+}
```
asoorm, line 129: Nice, can we make these functions a bit more generic? Lots of repetition here.

config/samples/httpbin.yaml
```diff
@@ -1,9 +1,12 @@
...
 metadata:
-  name: httpbin
+  name: httpbin2
 spec:
-  name: httpbin
+  name: httpbin2
+  contextRef: 
+    name: mycontext3
+    namespace: default
   use_keyless: true
```
asoorm, line 10: I wouldn't edit this particular sample, as this is the "hello-world" use-case.
Better to provide specific examples, such as here: `https://github.com/TykTechnologies/tyk-operator/tree/master/config/samples/07-operator-context`

controllers/apidefinition_controller.go
```diff
@@ -77,45 +78,60 @@ func (r *ApiDefinitionReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
 		if desired.Spec.OrgID == "" {
 			desired.Spec.OrgID = env.Org
 		}
+
 		util.AddFinalizer(desired, keys.ApiDefFinalizerName)
-		for _, certName := range desired.Spec.CertificateSecretNames {
+
+		// we support only one certificate secret name for mvp
+		if len(desired.Spec.CertificateSecretNames) != 0 {
+			certName := desired.Spec.CertificateSecretNames[0]
```
asoorm, line 114: What's this change about? is it a bug?

controllers/securitypolicy_controller.go
```diff
@@ -88,121 +92,160 @@ func (r *SecurityPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
 		return err
 	}
+
 	err = r.updateLinkedAPI(ctx, policy, func(ads *tykv1.ApiDefinitionStatus, s model.Target) {
 		ads.LinkedByPolicies = addTarget(ads.LinkedByPolicies, s)
 	})
+
+	if err != nil {
+		r.Log.Error(err, "failed to update linkedAPI status")
+	}
```
asoorm, line 242: If there is an error updating the linked API status - it's OK to continue? If so, why?

controllers/apidefinition_controller.go
komalsukhani: It was a for loop before which processed first record and then break unconditionally. Go linter was complaining about it.

config/samples/httpbin.yaml
komalsukhani: I modified it for some testing and forgot to undo changes. Reverted back.

integration/operator_context_test.go
```diff
@@ -0,0 +1,221 @@
...
+				err = client.Resources().Get(ctx, operatorCtx.Name, testNS, &opCtx)
+				if len(opCtx.Status.LinkedApiDefinitions) == 0 {
+					return errors.New("operator context status is not updated yet")
+				}
+
+				return nil
+			})
+			is.NoErr(err)
+
+			//time.Sleep(reconcileDelay)
```
komalsukhani, line 69: Tests are bit unstable now. Trying some retry logic. Once it fixes the issue I will remove this commented code.

integration/util.go
```diff
@@ -0,0 +1,29 @@
+package integration
+
+import (
+	"fmt"
+	"time"
+)
+
+func retryOperation(timeout time.Duration, interval time.Duration, fn func() error) error {
```
jlucktay, line 8: This could use a quick godoc to help make things clearer.
```suggestion
// retryOperation will retry the given 'fn' func at the given 'interval', until/unless 'timeout' is reached.
// Any non-nil errors returned by 'fn' in the interim will be ignored.
func retryOperation(timeout time.Duration, interval time.Duration, fn func() error) error {
```

integration/operator_context_test.go
```diff
@@ -0,0 +1,212 @@
...
+			is.NoErr(err) // failed to create operatorcontext
+
+			// create api definition
+			apiDef, err := createTestAPIDef(ctx, testNS, envConf)
+			is.NoErr(err) // failed to create apiDefinition
+
+			err = retryOperation(time.Hour, reconcileDelay, func() error {
+				var opCtx v1alpha1.OperatorContext
+				// shouldn't get deleted
+				err = client.Resources().Get(ctx, operatorCtx.Name, testNS, &opCtx)
```
jlucktay, line 60: I think these errors should be redeclared/renamed, to avoid confusion and/or shadowing.
```suggestion
			errRetry := retryOperation(time.Hour, reconcileDelay, func() error {
				var opCtx v1alpha1.OperatorContext
				// shouldn't get deleted
				errGet := client.Resources().Get(ctx, operatorCtx.Name, testNS, &opCtx)
```

jlucktay (commented):
Nice one, thanks @komalsukhani! 🙏 
I spotted a few different bits to clarify.

controllers/common.go
```diff
@@ -105,15 +106,117 @@ func httpContext(
...
+	}
+
+	switch object.(type) {
+	case *v1alpha1.ApiDefinition:
+		for _, opctx := range opCtxList.Items {
+			opctx.Status.RemoveLinkedAPIDefinition(target, log)
+
+			err := rClient.Status().Update(ctx, &opctx)
+			if err != nil {
+				log.Error(err, "Failed to remove link of APIDefintion from operator context", "operatorContext", opctx.Name, "apidefinition", target.Name)
```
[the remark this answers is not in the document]
komalsukhani: @asoorm Spoted the same issue.
I think we need to roll back all the operations performed before in case of error instead of returning directly. Otherwise it will leave the components in inconsistent state.

Didn't want to handle everything in single PR. But I could change it for now. Definitely need to be handled properly.

jlucktay (approved):
👏

PR #391, opened 2022-02-04
[TT-3700] Add JSON Schema Validation support

### LINKED ISSUE #59: mw: JSON Schema Validation support
Tyk Dashboard API expects the JSON Schema Validation schema to be presented as a JSON object.
Tyk Gateway API expects the JSON Schema Validation shema JSON to be presented as a b64 encoded string.

In this regard, the Tyk Operator with OSS Gateway will work, however because we cannot define arbitrary JSON `map[string]interface{}` as a Kubernetes CRD, it is currently not possible to support JSON Schema in a pro Tyk installation.

In this regard, one proposal would be to have an apidefinition_types.go and a dashboard_apidefinition_types, with a mediation layer to convert between them.

An alternative would be to have a custom marshaller & unmarshaller for the `map[string]interface{}`.

buraksekili:
## Description

This PR adds support for JSON Schema Validation in ApiDefinition.

## Related Issue

Resolves #59 

1. Create sample ApiDefinition that includes JSON Schema Validation, as follows:
    
     ```bash
    kubectl apply -f config/samples/httpbin_json_schema_validation.yaml
    ```
2. Make following requests to the gateway for testing schema validation as follows:
    1. The following one must fail because it does not meet schema's requirements (HTTP 422). 
     ```bash
     curl -i -X GET "http://localhost:8080/httpbin/get" -H "Content-Type: application/json" -d '{"userName":"b","age":0}'
     ```
    2. The following request must be successful (HTTP 200).
   
    
    ```bash
    curl -i -X GET "http://localhost:8080/httpbin/get" -H "Content-Type: application/json" -d '{"userName":"testuser","age":2}'
    ```
   

## Types of changes

## Checklist

  - [x] `gofmt -s -w .`
  - [x] `go vet ./...`
  - [x] `golangci-lint run`


integration/apidefinition_test.go
```diff
@@ -0,0 +1,150 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	"sigs.k8s.io/e2e-framework/klient/k8s"
+	"sigs.k8s.io/e2e-framework/klient/wait"
+	"sigs.k8s.io/e2e-framework/klient/wait/conditions"
+	"sigs.k8s.io/e2e-framework/pkg/envconf"
+	"sigs.k8s.io/e2e-framework/pkg/features"
+)
+
+const (
```
komalsukhani, line 24: I would suggest to set this as var type inside Test function itself. Since there will be more tests in this file which will not use these values.

integration/apidefinition_test.go
```diff
@@ -0,0 +1,150 @@
...
+				err := wait.For(conditions.New(client.Resources()).ResourceMatch(&desiredApiDef, func(object k8s.Object) bool {
+					apiDef := object.(*v1alpha1.ApiDefinition) //nolint:errcheck
+					// 'validate_json' field must exist in the ApiDefinition object.
+					return len(apiDef.Spec.VersionData.Versions[defaultVersion].ExtendedPaths.ValidateJSON) == 1
+				}), wait.WithTimeout(defaultTimeout))
+				is.NoErr(err)
+
+				return ctx
+			}).
+		Assess("ApiDefinition must exists in the Gateway",
```
komalsukhani, line 93: Since you are sending request to API in the next assessment can this be removed?

integration/apidefinition_test.go
buraksekili: 👍 I moved them to the inside of the Test function.

api/model/api.go
```diff
@@ -318,14 +318,21 @@ type MethodTransformMeta struct {
...
-type ValidatePathMeta struct {
-	// Allows override of default 422 Unprocessible Entity response code for validation errors.
-	ErrorResponseCode int        `json:"error_response_code"`
-	Path              string     `json:"path"`
-	Method            HttpMethod `json:"method"`
-	SchemaB64         string     `json:"schema_b64,omitempty"`
+// JSONValidationSchema represents 'schema' field of the 'validate_json'.
+// Validate JSON verifies user requests against a specified JSON schema and
+// check that the data sent to your API by a consumer is in the right format.
+type JSONValidationSchema struct {
```
komalsukhani, line 324: Instead of defining new type here, we can reuse the `ConfigDataType`. Just need to rename it to something generic.

integration/apidefinition_test.go
```diff
@@ -0,0 +1,129 @@
...
+func TestApiDefinitionCreate(t *testing.T) {
+	var (
+		apiDefWithJSONValidationName = "apidef-json-validation"
+		apiDefListenPath             = "/validation"
+		defaultVersion               = "Default"
+		errorResponseCode            = 422
+		defaultTimeout               = 1 * time.Minute
+	)
+
+	eps := &model.ExtendedPathsSet{
```
andrei-tyk, line 33: shouldn't we isolate this into a subtest so that for example when we add unit tests for each type of extenedPath we can just write another subtests instead of a whole test?

api/model/api.go
buraksekili: Okay @komalsukhani 👍 , what about `UnstructuredField`? Do you have any suggestions on your mind regarding the naming?

komalsukhani: @buraksekili It could be `MapStringInterfaceType`. Since we are using it specifically for that type.

buraksekili: makes sense 👍 , i'll update the PR soon

Makefile
```diff
@@ -40,7 +40,7 @@ UNIT_TEST=$(shell go list ./... | grep -v bdd)
 
 test: generate fmt vet manifests
 	go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
-	setup-envtest use
+	setup-envtest --arch=amd64 use
```
andrei-tyk, line 43: I do not think this m1 hack should be left alone on the repo, maybe make an extra Makefile target "test-m1" with this

buraksekili: If `--arch=amd64` causes confusion in the `test` command, we can separate it. However, I do not believe that we really need to have 2 commands for the same purposes since amd64-based architectures are using `setup-envtest` with the `--arch=amd64` option by default https://github.com/kubernetes-sigs/controller-runtime/blob/b1d9a1641f312ed8b526506486415a3983d4e6d1/tools/setup-envtest/main.go#L43.

api/model/api.go
```diff
@@ -319,13 +322,14 @@ type MethodTransformMeta struct {
 }
 
 type ValidatePathMeta struct {
-	// Allows override of default 422 Unprocessible Entity response code for validation errors.
+	// Allows override of default 422 Unprocessable Entity response code for validation errors.
 	ErrorResponseCode int        `json:"error_response_code"`
 	Path              string     `json:"path"`
 	Method            HttpMethod `json:"method"`
 	SchemaB64         string     `json:"schema_b64,omitempty"`
```
buraksekili, line 329: @asoorm  Should we continue to support the SchemaB64 version after enabling Schema? Because the ValidateJSON field of ExtendedPathsSets was previously commented, now I uncommented it.

cc: @komalsukhani

asoorm: @buraksekili I may be wrong, but I seem to remember that we support both because one version is required for tyk with dashboard, and another one for Tyk CE, or maybe the Dashboard stores the payload B64 encoded inside Mongo. 

It's certainly worth double checking with @furkansenharputlu and @buger what makes sense and how this logic works from the dashboard and Gateway side. 

My initial thoughts are that we might need to support both inside the K8s CRD and leave it to the end-user to decide what they want to send us. Then the reconcile loop would be responsible for transforming the payload as required for the Gateway or Dashboard client. The reason for these initial thoughts is because in a dev environment, I would want to be able to export my API Definition AS IS (and regardless of whether CE or PRO), and I would want to be able to drop that in to a CRD without having to do any manual conversion.

buraksekili: Thanks @asoorm, I will leave the `SchemaB64` as it is.

buraksekili: Since providing `schema` is enough for validating JSON, I removed `schema_b64` from ValidatePathMeta.

komalsukhani (approved):

komalsukhani:
Thank you @buraksekili !!

PR #394, opened 2022-02-10
[TT-4489] Documentation on current UDG support

buraksekili:
## Description
This PR adds documentation on current UDG support in the Operator, updates validation webhooks for ApiDefinition and fixes the UDG example YAML.

## Motivation and Context

1. Previously, it was not feasible to create an ApiDefinition with an empty `target_url` field due to validation. Therefore, the validation webhook has been updated to allow UDG users to define an empty target URL. 
2. Also, this PR updates the ApiDefinition CRD because the previous CRD omits empty target URLs while making requests to your Tyk installation. Therefore, the ApiDefinitionSpec is updated to prevent omitting empty `target_urls`.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] My change requires a change to the documentation.
  - [x] `gofmt -s -w .`
  - [x] `go vet ./...`
  - [x] `golangci-lint run`

### DIFF
--- a/api/v1alpha1/apidefinition_webhook_test.go
+++ b/api/v1alpha1/apidefinition_webhook_test.go
@@ -143,8 +143,32 @@ func TestApiDefinition_validateTarget(t *testing.T) {
 			expectedErr: nil,
 		},
 		{
+			name: "valid ApiDefinition, targeting internal API",
+			apiDef: ApiDefinition{
+				Spec: APIDefinitionSpec{
+					APIDefinitionSpec: model.APIDefinitionSpec{
+						Proxy: model.Proxy{TargetURL: "", TargetInternal: &model.TargetInternal{
+							Target: model.Target{Name: "resource-name", Namespace: "resource-ns"},
+						}},
+					},
+				},
+			},
+			expectedErr: nil,
+		},
+		{
 			name:        "invalid ApiDefinition, missing proxy.target_url",
 			apiDef:      ApiDefinition{},
+			expectedErr: &missingTargetURLErr,
+		},
+		{
+			name: "invalid ApiDefinition, empty proxy.target_url",
+			apiDef: ApiDefinition{
+				Spec: APIDefinitionSpec{
+					APIDefinitionSpec: model.APIDefinitionSpec{
+						Proxy: model.Proxy{TargetURL: ""},
+					},
+				},
+			},
 			expectedErr: &missingTargetURLErr,
 		},
 		{


api/v1alpha1/apidefinition_webhook.go
```diff
@@ -202,14 +202,24 @@ func (in *ApiDefinition) ValidateDelete() error {
...
-	if in.Spec.Proxy.TargetURL == "" && in.Spec.Proxy.TargeInternal == nil {
-		all = append(all,
-			field.Required(path("proxy", "target_url"),
-				"can't be emptry",
-			),
-		)
+	// TargetURL is only allowed to be an empty string when we have GraphQL
+	// API, or we are targeting internal API by its name and namespace.
+	if in.Spec.Proxy.TargetURL == "" {
+		if in.Spec.GraphQL != nil {
```
komalsukhani, line 208: We can merge both the checks together
```suggestion
		if in.Spec.GraphQL != nil && !in.Spec.GraphQL.Enabled {
```

api/v1alpha1/apidefinition_webhook.go
```diff
@@ -202,14 +202,24 @@ func (in *ApiDefinition) ValidateDelete() error {
 
 func (in *ApiDefinition) validateTarget() field.ErrorList {
```
asoorm, line 203: Due to the relatively high cyclomatic complexity of this function, i think it might be a good idea to write a test covering each scenario.

buraksekili: I've added test cases for `validateTarget`

api/v1alpha1/apidefinition_webhook_test.go
```diff
@@ -80,3 +81,125 @@ func TestApiDefinition_Default_DoNotTrack(t *testing.T) {
...
+				Spec: APIDefinitionSpec{
+					APIDefinitionSpec: model.APIDefinitionSpec{
+						Proxy: model.Proxy{TargetURL: "/test"},
+					},
+				},
+			},
+			expectedErr: nil,
+		},
+		{
+			name:        "invalid ApiDefinition, missing proxy.target_url",
```
komalsukhani, line 146: @buraksekili Can you add a test where `TargetURL` is set to "" ?
Also when `TargetInternal` is set and `TargetURL` is not.

buraksekili: Thanks @komalsukhani. I added new test cases to cover the scenarios that you mentioned.

komalsukhani (approved):

PR #415, opened 2022-03-22
[TT-4460] - Add support of the certificate pinning

### LINKED ISSUE #405: [TT-4460] - certificate pinning
There is a request for supporting Certificate Pinning via Operator

buraksekili:
## Description
This PR adds the support of `pinned_public_keys` field, which is used for [Certificate Pinning](https://tyk.io/docs/security/certificate-pinning/), in the CRD.

## Related Issue
Resolves #405 

_Please see the example manifest (`httpbin_certificate_pinning.yaml`) attached to this PR for reference._

## Types of changes

## Checklist

- [x] Make sure you are updating [CHANGELOG.md](../CHANGELOG.md) based on your changes.
  - [x] `gofmt -s -w .`
  - [x] `go vet ./...`
  - [x] `golangci-lint run`


andrei-tyk (changes requested):

integration/apidefinition_test.go
```diff
@@ -508,3 +508,59 @@ func TestApiDefinitionCreateIgnored(t *testing.T) {
...
+			// Create ApiDefinition with Certificate Pinning.
+			_, err := createTestAPIDef(ctx, testNS, func(apiDef *v1alpha1.ApiDefinition) {
+				apiDef.Name = apiDefCertificatePinningName
+				apiDef.Spec.PinnedPublicKeys = map[string]string{"*": publicKeyID}
+			}, envConf)
+			is.NoErr(err) // failed to create apiDefinition
+
+			return ctx
+		}).
+		Assess("ApiDefinition must have Certificate Pinning field defined",
```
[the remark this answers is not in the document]
buraksekili: @andrei-tyk Added new test cases 👍

andrei-tyk (approved):

controllers/secret_cert_controller.go
```diff
@@ -104,28 +107,58 @@ func (r *SecretCertReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+			return ctrl.Result{}, nil
+		}
+	case opaqueSecretType:
+		var (
+			ok                        bool
+			publicKeyFieldName        = "public-key"
+			publicKeyEnabledFieldName = "public-key-enabled"
+		)
+		// public-key-enabled field indicates that this resource will be tracked by the secret controller.
+		enabled, exists := desired.Data[publicKeyEnabledFieldName]
```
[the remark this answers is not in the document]
buraksekili: Since we do not have certain field names to store public key (for example, kubernete.io/tls types use `tls.crt` & `tls.key`), we are getting a field name where the cert-controller should check for public key. If users provide an invalid field name, the secret object will not have the public key in that field. Therefore, i've thought that we should check the existence of the public key.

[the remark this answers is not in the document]
andrei-tyk: If "public-key" is not a standard field for a secret resource, couldn't we just check that the public key exists?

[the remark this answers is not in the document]
buraksekili: Yes `public-key` is not standard field for a secret resource. I just want to prevent checking all Opaque secrets by secret-controller via checking `enabled` field. If users provide `enabled` field, secret-controller checks that secret. Otherwise, it is not managed by cert-controller as before. If you think it is not necessary, we can remove that part.

andrei-tyk (approved):

api/model/api.go
```diff
@@ -508,6 +508,16 @@ type OpenIDOptions struct {
...
 }
 
+// PinnedPublicKeySecret holds information about a Kubernetes Secret object called 'SecretName' residing in
+// 'SecretNamespace' namespace that includes the public key of the domain, defined through 'PublicKeySecretField' field.
+type PinnedPublicKeySecret struct {
+	SecretName string `json:"secret_name"`
+
+	// +optional
+	// +kubebuilder:default:=default
+	SecretNamespace string `json:"secret_namespace,omitempty"`
```
andrei-tyk, line 518: The namespace should be the same as the one used in the ApiDefinition => the check should be done there.

andrei-tyk (changes requested):

config/samples/httpbin_certificate_pinning.yaml
```diff
@@ -0,0 +1,65 @@
...
+# Now, in the ApiDefinition, we are using this public key for all domains as '*' (wildcard domains are also supported).
+# If you target any other URL than https://httpbin.org (for instance https://github.com/), you will face a 'public key pinning error'
+# for that particular domain because the public key of httpbin.org is used for all domains.
+apiVersion: v1
+kind: Secret
+metadata:
+  name: httpbin-secret
+type: Opaque
+data:
+  public-key: | # This field name is fixed and shouldn't be changed. Secret controller strictly checks for `public-key` field.
```
asoorm, line 42: Is that a good design idea?

controllers/apidefinition_controller.go
```diff
@@ -132,29 +132,54 @@ func (r *ApiDefinitionReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
 
+		// Check Pinned Public keys
+		if len(desired.Spec.PinnedPublicKeysSecretNames) != 0 {
+			if upstreamRequestStruct.Spec.PinnedPublicKeys == nil {
+				upstreamRequestStruct.Spec.PinnedPublicKeys = map[string]string{}
+			}
+
+			for domain, secretName := range desired.Spec.PinnedPublicKeysSecretNames {
+				// Set the namespace for referenced secret to the current namespace where ApiDefinition lives.
+				tykCertID, err := r.checkSecretAndUpload(ctx, secretName, req.Namespace, log, &env, "public-key")
```
asoorm, line 151: if `public-key` is static, then it should be a constant. But I don't know if it is a good idea to be explicit about it - e.g. a public key generated by cert-manager might be called something like `tls.crt`. Is it worth considering how we can be a bit more interoperable with standard tooling?

config/samples/httpbin_certificate_pinning.yaml
komalsukhani: @asoorm Isn't tls secret used to store private key and certificate?

controllers/apidefinition_controller.go
```diff
@@ -132,29 +132,54 @@ func (r *ApiDefinitionReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+				if err != nil {
+					// we should log the missing secret, but we should still create the API definition
+					log.Info(
```
komalsukhani, line 154: Shouldn't it be an error?

controllers/apidefinition_controller.go
```diff
@@ -172,49 +197,71 @@ func (r *ApiDefinitionReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+		}
+
+		return uploadCert(ctx, env.Org, pemKeyBytes, pemCrtBytes)
 	}
 
-	pemKeyBytes, ok := secret.Data["tls.key"]
+	pubKey, ok := secret.Data[field]
 	if !ok {
-		err = fmt.Errorf("%s", "requeueing because key not found in secret")
+		err = fmt.Errorf("requeueing because %s not found in secret", field)
```
komalsukhani, line 258: I think we should just log error here. Requeueing or not should be left to the function calling this function.

controllers/secret_cert_controller.go
```diff
@@ -177,7 +198,44 @@ func (r *SecretCertReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+					apiDefList.Items[idx].Spec.PinnedPublicKeys = map[string]string{}
+				}
+
+				apiDefList.Items[idx].Spec.PinnedPublicKeys[domain] = certID
+
+				err = r.Update(ctx, &apiDefList.Items[idx])
+
+				if apierrors.IsConflict(err) {
+					// The Pod has been updated since we read it.
+					// Requeue the Pod to try to reconciliate again.
```
komalsukhani, line 224:
```suggestion
					// Requeue to try to reconciliate again.
```

controllers/secret_cert_controller.go
```diff
@@ -177,7 +198,44 @@ func (r *SecretCertReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+					return ctrl.Result{Requeue: true}, nil
+				}
+
+				if apierrors.IsNotFound(err) {
+					// The Pod has been deleted since we read it.
+					// Requeue the Pod to try to reconciliate again.
```
komalsukhani, line 230:
```suggestion
					// Requeue to try to reconciliate again.
```

docs/api_definitions.md
```diff
@@ -54,11 +54,11 @@ An API Definition describes the configuration of an API. It instructs Tyk Gatewa
...
 
-| Type | Supported | Comments |
-| ----------- | --------- | --------- |
-| Public Key Certificate Pinning | ❌ | Not implemented |
-| Upstream Certificates mTLS | ✅ | Not implemented |
-| Request Signing | ❌ | Not implemented |
+| Type                                                                                 | Supported | Comments        |
+|--------------------------------------------------------------------------------------|-----------|-----------------|
+| [Public Key Certificate Pinning](../config/samples/httpbin_certificate_pinning.yaml) | ✅         |                 |
+| Upstream Certificates mTLS                                                           | ✅          | Not implemented |
```
komalsukhani, line 60: It says not implemented in the comments

pkg/cert/cert.go
```diff
@@ -12,6 +12,7 @@ func HexSHA256(cert []byte) string {
 }
 
 func CalculateFingerPrint(data []byte) string {
+	// TODO: publicKey might be returned as nil. if that's the case, the program panics.
```
komalsukhani, line 15: Do you want to handle it in this PR?

controllers/apidefinition_controller.go
buraksekili: If users target missing secret, we are creating the ApiDefinition as it is except Public Key Pinning feature. Once the secret is created, secret_cert controller creates the certificate and update the ApiDefinition's Public Key Pinning field. This is the same flow in the Upstream Certificate https://github.com/TykTechnologies/tyk-operator/blob/master/controllers/apidefinition_controller.go#L116

controllers/apidefinition_controller.go
```diff
@@ -132,29 +132,54 @@ func (r *ApiDefinitionReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+				upstreamRequestStruct.Spec.PinnedPublicKeys = map[string]string{}
+			}
+
+			for domain, secretName := range desired.Spec.PinnedPublicKeysSecretNames {
+				// Set the namespace for referenced secret to the current namespace where ApiDefinition lives.
+				tykCertID, err := r.checkSecretAndUpload(ctx, secretName, req.Namespace, log, &env, "public-key")
+				if err != nil {
+					// we should log the missing secret, but we should still create the API definition
+					log.Info(
+						fmt.Sprintf("cert %s/%s is missing", secretName, req.Namespace),
```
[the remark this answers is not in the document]
buraksekili: There are two possible error scenarios: errors regarding missing secret and errors happened while uploading the certifcate to Tyk. I will update the error message as you suggested 👍

controllers/apidefinition_controller.go
```diff
@@ -172,49 +197,71 @@ func (r *ApiDefinitionReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
 
-	pemCrtBytes, ok := secret.Data["tls.crt"]
-	if !ok {
-		err = fmt.Errorf("%s", "requeueing because cert not found in secret")
-		log.Error(err, "requeueing because cert not found in secret")
+	// 'field' holds Secret object's data field that we are going to check.
+	// If kubernetes.io/tls type Secret is used, there is no need to provide 'field' since kubernetes.io/tls types have
+	// already defined fields called 'tls.crt' and 'tls.key'. However, if you provide any other type of secrets such
+	// as Opaque (as we do in Public Key Certificate pinning), you need to provide field that controller should check,
+	// through 'public_key_secret_field'.
```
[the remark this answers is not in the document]
buraksekili: Ups, it is deprecated. I forgot to update it. will update 👍

pkg/cert/cert.go
buraksekili: It'd be better if we handle it in a separate PR because `CaclulateFingerPrint` function is called in multiple places.

config/samples/httpbin_certificate_pinning.yaml
buraksekili: For future reference, we decided to use the tls-secret type instead of Opaque

controllers/apidefinition_controller.go
buraksekili: This part becomes deprecated once we remove Opaque support. Therefore, I kept the logic here as it is

docs/api_definitions.md
buraksekili: updated 👍

controllers/secret_cert_controller.go
```diff
@@ -177,7 +177,44 @@ func (r *SecretCertReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+				err = r.Update(ctx, &apiDefList.Items[idx])
+
+				if apierrors.IsConflict(err) {
+					// The Pod has been updated since we read it.
+					// Requeue to try to reconciliate again.
+					return ctrl.Result{Requeue: true}, nil
+				}
+
+				if apierrors.IsNotFound(err) {
+					// The Pod has been deleted since we read it.
```
komalsukhani, line 208:
```suggestion
					// The Pod has been deleted since we read it.
```

Should it be APIDefiniton/resource instead of Pod?

controllers/secret_cert_controller.go
```diff
@@ -177,7 +177,44 @@ func (r *SecretCertReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+					// Requeue to try to reconciliate again.
+					return ctrl.Result{Requeue: true}, nil
+				}
+
+				if err != nil {
+					log.Error(err, "unable to update ApiDef")
```
komalsukhani, line 214: Should we return an err here too?
Otherwise it will print next log line that API Def was updated successfully.

buraksekili: I've updated this part of the code. So, it will return err right now.

andrei-tyk (approved):

PR #417, opened 2022-03-22
[TT-4598] Added support of new fields for GraphQL ProxyOnly API

komalsukhani:
## Description

- Added support of `proxy.auth_header` fields for GraphQL ProxyOnly API. 
If upstream server is protected, it can be used to pass Authorization details to the upstream.
- Added documentation

## Related Issue

[TT-4598](https://tyktech.atlassian.net/browse/TT-4598)

- I have tested it locally(Haven't added sample manifest, as it will require Authorization details to work)
- Have added integration tests for graphql API
- Have added unit test for api definition webhook to improve test coverage of PR.

## Types of changes

## Checklist

- [x] Make sure you are updating [CHANGELOG.md](../CHANGELOG.md) based on your changes.
  - [x] `gofmt -s -w .`
  - [x] `go vet ./...`
  - [x] `golangci-lint run`


api/model/api.go
```diff
@@ -983,6 +982,14 @@ type GraphQLConfig struct {
...
 	// GraphQLPlayground is the Playground specific configuration.
 	GraphQLPlayground GraphQLPlayground `json:"playground,omitempty"`
+
+	// Proxy holds the configuration for a proxy only api.
+	Proxy GraphQLProxyConfig `bson:"proxy" json:"proxy,omitempty"`
+}
+
+type GraphQLProxyConfig struct {
+	// +nullable
+	AuthHeaders map[string]string `bson:"auth_headers" json:"auth_headers,omitempty"`
```
buraksekili, line 992: Once users use `proxy` field within their `GraphQLConfig`, shouldn't they need to provide `AuthHeaders`? If not, is it okay to provide `AuthHeaders` as null when `Proxy` is declared?

Also, do we need to use `bson` struct tag?

komalsukhani: You are right. AuthHeaders should not be nullable. 
I will fix that and also remove bson tag.

asoorm: What is different about this field in comparison with a more generic option such as UpstreamHeaders? e.g. does it automatically inject "Bearer " prefix? Why would you need multiple auth headers?

It would be good to document this difference (if any) so that the docs propagate to the CRD. Maybe i'm looking too much into it, but im not sure why it would be named as-such.

komalsukhani: Screenshot from 2022-03-23 17-27-18
@asoorm Are you referring to `Upstream Auth Headers` field on this screen?
If you are, it sets `proxy.auth_headers` field in the APIDefinition

komalsukhani: @buraksekili If I remove nullable for AuthHeader, API Definition creation fails if AuthHeaders aren't specified with error stating "Object can't have null value". Have to add it back

api/v1alpha1/apidefinition_webhook_test.go
```diff
@@ -239,3 +242,300 @@ func hasError(errs field.ErrorList, needle string) bool {
 
 	return false
 }
+
+func TestApiDefinition_Validate_Auth(t *testing.T) {
+	is := is.New(t)
+
+	tests := map[string]struct {
+		ApiDefinition ApiDefinition
+		ReturnErr     bool
```
buraksekili, line 251: Do we need `ReturnErr` field to be present in the test case struct? Can't we just compare `ErrCause` with the `err` returned by `ApiDefinition.validate()`?

andrei-tyk (approved):

PR #419, opened 2022-03-23
[TT-4925] Fixed wrongly storing mutated CRDs in kubernetes instead of the initi…

andrei-tyk:
Fixed wrongly storing mutated CRDs in kubernetes instead of the initial ones by using a deepcopy that is mutated and then sent to the gateway.

Due to using a single ApiDefinition struct for storing both the “input” and “output” CRD values, at some point if we want to work with values that are not going to be passed to the gateway/dashboard (as they would fail the api structure validator) we used to mutate the ApiDefinitions obtained from kubernetes directly. This behaviour led to the kubernetes API being unable to recognize whether a CRD was created or updated (as the stored CRD would be different from the one that was applied initially).

## Description

A solution identified together with Ahmet Soormally relies on deep-copying the initial ApiDefinition spec into an additional variable and mutating that and sending it to the gateway.

## Motivation and Context

This behaviour led to the kubernetes API being unable to recognize whether a CRD was created or updated (as the stored CRD would be different from the one that was applied initially).

## Types of changes

## Checklist

  - [ ] `gofmt -s -w .`
  - [ ] `go vet ./...`
  - [ ] `golangci-lint run`

### DIFF
--- a/controllers/apidefinition_controller.go
+++ b/controllers/apidefinition_controller.go
@@ -116,9 +116,9 @@ func (r *ApiDefinitionReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		// we support only one certificate secret name for mvp
 		if len(desired.Spec.CertificateSecretNames) != 0 {
 			certName := desired.Spec.CertificateSecretNames[0]
-			tykCertID, err2, done := r.checkSecretAndUpload(ctx, certName, namespacedName, log, &env)
-			if done {
-				return err2
+			tykCertID, err, shouldRequeue := r.checkSecretAndUpload(ctx, certName, namespacedName, log, &env)
+			if shouldRequeue {
+				return err
 			}
 
 			upstreamRequestStruct.Spec.Certificates = []string{tykCertID}
--- a/controllers/apidefinition_controller.go
+++ b/controllers/apidefinition_controller.go
@@ -116,8 +116,8 @@ func (r *ApiDefinitionReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		// we support only one certificate secret name for mvp
 		if len(desired.Spec.CertificateSecretNames) != 0 {
 			certName := desired.Spec.CertificateSecretNames[0]
-			tykCertID, err, shouldRequeue := r.checkSecretAndUpload(ctx, certName, namespacedName, log, &env)
-			if shouldRequeue {
+			tykCertID, err := r.checkSecretAndUpload(ctx, certName, namespacedName, log, &env)
+			if err != nil {
 				return err
 			}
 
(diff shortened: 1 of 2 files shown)


buraksekili (approved):

controllers/apidefinition_controller.go
```diff
@@ -180,6 +156,50 @@ func (r *ApiDefinitionReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+	err := r.Get(ctx, types.NamespacedName{Name: certName, Namespace: namespacedName.Namespace}, &secret)
+	if err != nil {
+		log.Error(err, "requeueing because secret not found")
+
+		return "", err, true
+	}
+
+	pemCrtBytes, ok := secret.Data["tls.crt"]
+	if !ok {
+		log.Error(err, "requeueing because cert not found in secret")
```
komalsukhani, line 177: `err` would be nil in this case

controllers/apidefinition_controller.go
```diff
@@ -100,56 +104,28 @@ func (r *ApiDefinitionReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
-				log.Error(err, "requeueing because secret not found")
-
-				return err
-			}
-
-			pemCrtBytes, ok := secret.Data["tls.crt"]
-			if !ok {
-				log.Error(err, "requeueing because cert not found in secret")
-
+			tykCertID, err, shouldRequeue := r.checkSecretAndUpload(ctx, certName, namespacedName, log, &env)
```
komalsukhani, line 119: Why do we need bool here? Can't we check err value here?

controllers/apidefinition_controller.go
```diff
@@ -180,6 +156,50 @@ func (r *ApiDefinitionReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+	pemCrtBytes, ok := secret.Data["tls.crt"]
+	if !ok {
+		log.Error(err, "requeueing because cert not found in secret")
+
+		return "", err, true
+	}
+
+	pemKeyBytes, ok := secret.Data["tls.key"]
+	if !ok {
+		log.Error(err, "requeueing because key not found in secret")
```
asoorm, line 184: We need to log an error, but `err` is nil, because `pemKeyBytes, ok := secret.Data["tls.key"]` doesn't set err

andrei-tyk: I have updated the PR with the required changes.

controllers/apidefinition_controller.go
andrei-tyk: I have updated the PR with the required changes.

komalsukhani (approved):

PR #420, opened 2022-03-24
[TT-5067] Build process changes

komalsukhani:
## Description

- **Preloading of images is turned off by default.** 
      Added `TYK_OPERATOR_PRELOAD_IMAGES` environment variable to enable/disable preload images step. Skipping preloading of images step has reduced setup time on GithubActions by approx 2m.
This step was also causing issue on Mac machine. Now Mac users can use Makefile commands to create their setup.
- Changed deployment type of gateway in Pro mode from `DaemonSet` to `Deployment`.

## Related Issue

[TT-5067](https://tyktech.atlassian.net/browse/TT-5067)

## Motivation and Context

Preloading of images wasn't adding any advantage in GithubActions. By default it is turned off. Can be turned on by setting `TYK_OPERATOR_PRELOAD_IMAGES` environment variable to true.

## Types of changes

## Checklist

  - [x] `gofmt -s -w .`
  - [x] `go vet ./...`
  - [x] `golangci-lint run`


buraksekili (approved):

buraksekili:
It looks good to me but i have a question. I am not sure but I would like to ask something about deployment type of gateway. Since we have 2 worker nodes in our development environment, shouldn't we use Daemonset instead of Deployments?
> A DaemonSet ensures that all (or some) Nodes run a copy of a Pod. As nodes are added to the cluster, Pods are added to them. As nodes are removed from the cluster, those Pods are garbage collected. Deleting a DaemonSet will clean up the Pods it created.
Ref: https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/

What advantages does using Deployment over Daemonset bring?

komalsukhani:
> It looks good to me but i have a question. I am not sure but I would like to ask something about deployment type of gateway. Since we have 2 worker nodes in our development environment, shouldn't we use Daemonset instead of Deployments?
> 
> > A DaemonSet ensures that all (or some) Nodes run a copy of a Pod. As nodes are added to the cluster, Pods are added to them. As nodes are removed from the cluster, those Pods are garbage collected. Deleting a DaemonSet will clean up the Pods it created.
> > Ref: https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/
> 
> What advantages does using Deployment over Daemonset bring?

For production you need to have Gateway running on each node.
But for testing purpose, having just one instance of Gateway should be enough according to me.

PR #458, opened 2022-05-17
[TT-5320] - Support different namespaces for SubGraph

buraksekili:
## Description
This PR adds a support for subgraphs to be created in different namespaces; and then joint together by a single supergraph.

## Motivation and Context

The current implementation of federation support on the operator does not allow users to use subgraphs in different namespaces than supergraph. So, if the supergraph is in the `default` namespace, Operator was not able to find referenced subgraphs which are not in `default` namespace.

This PR aims to support use of subgraphs in different namespaces in the SuperGraph CRD.
For example, let's say you have `Users` SubGraph in `users-ns` namespace.
This SubGraph must be accessible for SuperGraph as well. So that we can give reference of it in SuperGraph CRD declaration.

## Types of changes

## Checklist

  - [x] `gofmt -s -w .`
  - [x] `go vet ./...`
  - [x] `golangci-lint run`


helm/crds/crds.yaml
```diff
@@ -3182,7 +3182,15 @@ spec:
                 type: string
               subgraphs_refs:
```
asoorm, line 3183: this should be `subgraph_refs`

api/v1alpha1/supergraph_types.go
```diff
@@ -20,11 +20,16 @@ import (
...
+	Name      string `json:"name"`
+	Namespace string `json:"namespace"`
+}
+
 // SuperGraphSpec defines the desired state of SuperGraph
 type SuperGraphSpec struct {
-	SubgraphsRefs []string `json:"subgraphs_refs"`
-	MergedSDL     string   `json:"merged_sdl,omitempty"`
-	Schema        string   `json:"schema,omitempty"`
+	SubgraphsRefs []SubGraphReference `json:"subgraphs_refs"`
```
asoorm, line 30: Don't we already have a more generic container type for this which is used elsewhere? e.g. something like `type NamespacedName struct {...`?

asoorm (changes requested):

buraksekili:
closed in favor of #466

PR #481, opened 2022-07-14
CONCEPT: CLI to pull APIs from an existing Tyk installation and convert to K8s YAML format

asoorm:
Concept


pkg/snapshot/README.md
```diff
@@ -0,0 +1,188 @@
...
+This tool is provided as PoC only and bear some limitations and restrictions. Please use with caution.
+
+| [Pre-requisite](#pre-requisite) | [Installation](#installation) | [Usage](#usage) | [Limitations](#limitations) |
+
+## Pre-Requisite
+
+1. Access to `tyk-operator` repository.
+2. Credentials to connect Tyk Dashboard. Please visit [Tyk Docs](https://tyk.io/docs/tyk-stack/tyk-operator/installing-tyk-operator/#tyk-self-managed-hybrid)
+for details.
+
```
caroltyk, line 19: Include [go](https://go.dev/doc/install) too?

pkg/snapshot/README.md
```diff
@@ -0,0 +1,188 @@
...
+cd tyk-operator/
+git checkout feat/importer-cli
+```
+
+2. Build Tyk Operator
+```bash
+go build
+```
+
+3. Run snapshot
```
caroltyk, line 37: I think we can move this step to "Usage"

api/model/api.go
```diff
@@ -60,7 +60,7 @@ type (
 type HttpMethod string
 
 // GraphQLExecutionMode is the mode to define how an api behaves.
-// +kubebuilder:validation:Enum=proxyOnly;executionEngine;supergraph;subgraph
+// +kubebuilder:validation:Enum="";proxyOnly;executionEngine;supergraph;subgraph
```
asoorm, line 63: Could you explain why we are allowing empty value here for GraphQLExecutionMode?

api/model/api.go
```diff
@@ -69,7 +69,7 @@ const (
 )
 
 // APIProtocol is the network transport protocol supported by the gateway
-// +kubebuilder:validation:Enum=h2c;tcp;tls;http;https;
+// +kubebuilder:validation:Enum="";h2c;tcp;tls;http;https;
```
asoorm, line 72: Could you explain why we are permitting empty value here for APIProtocol?

pkg/snapshot/README.md
```diff
@@ -0,0 +1,203 @@
...
+
+
+### Exporting API Definitions
+
+#### Specify Category to export
+
+By default, tyk-operator exports ApiDefinitions defined in `#operator` category. You can change default category via `--category` flag.
+
+```bash
+TYK_MODE=pro TYK_URL=<TYK_URL> TYK_AUTH=<TYK_AUTH> TYK_ORG=<TYK_ORG> ./tyk-operator --snapshot output.yaml --category k8s
```
asoorm, line 99: Rather than `<TYK_URL>`, I would prefer `${TYK_URL}` that way the snippet can be copied and pasted - and it would just work if we have an env var.

pkg/snapshot/README.md
```diff
@@ -0,0 +1,203 @@
...
+```
+
+You can also dump all ApiDefinitions from Dashboard via `--all` flag.
+
+```bash
+TYK_MODE=pro TYK_URL=<TYK_URL> TYK_AUTH=<TYK_AUTH> TYK_ORG=<TYK_ORG> ./tyk-operator --snapshot output.yaml --all
```
asoorm, line 105: Rather than `<TYK_URL>`, I would prefer `${TYK_URL}` that way the snippet can be copied and pasted - and it would just work if we have an env var.

pkg/snapshot/README.md
```diff
@@ -0,0 +1,203 @@
...
+  ...
+```
+
+In order to specify CR metadata, you can use Config Data. For specified ApiDefinitions,
+snapshot CLI generates ApiDefinition CRs based on Config Data of that specific ApiDefinition.
+
+```json
+{
+  "k8sName": "metadata-name",
+  "k8sNs": "metadata-namespace"
```
asoorm, line 129: I would suggest that `k8sNs` should be renamed to `k8sNamespace`.
I would also recommend that if empty - it does not pick `default` but rather allows the user to choose when they kubectl apply.

e.g. `kubectl apply -f ./foo.yaml -n bar` in order to target the `bar` namespace.

pkg/snapshot/README.md
```diff
@@ -0,0 +1,203 @@
...
+
+The CLI checks for `k8sName` and `k8sNs` fields of each ApiDefinition's Config Data 
+to generate metadata of the output CR. If these fields exist, the CLI uses the 
+values specified in these fields. Otherwise, it uses default values (`replace-me-` 
+for name and `default` for namespace) for them.
+
+Assume we have the following ApiDefinitions, two of which are categorized as `#testing` 
+and created on our Dashboard.
+
+![apis](./img/apis.png)
```
asoorm, line 141: please provide a text equivalent / alt text of screenshot.

pkg/snapshot/README.md
```diff
@@ -0,0 +1,203 @@
...
+
+If we would like to specify metadata of the `test-api-5`, we can update Config Data
+of the ApiDefinition, as follows.
+
+![config-data](./img/config-data.png)
```
asoorm, line 146: please provide text equivalent of config-data screenshot.

pkg/snapshot/README.md
```diff
@@ -0,0 +1,203 @@
...
+
+So, the generated output for this environment will look as follows;
+```bash
+TYK_MODE=pro TYK_URL=<TYK_URL> TYK_AUTH=<TYK_AUTH> TYK_ORG=<TYK_ORG> ./tyk-operator --snapshot output.yaml --category testing
```
asoorm, line 150: Prefer `${TYK_URL}`, `${TYK_AUTH}`... etc

pkg/snapshot/README.md
```diff
@@ -0,0 +1,203 @@
...
+spec:
+  name: 'test-api-5 #testing'
+  ...
+---
+apiVersion: tyk.tyk.io/v1alpha1
+kind: ApiDefinition
+metadata:
+  creationTimestamp: null
+  name: replace-me-1    # Since Config Data does not include "k8sName", default name is used.
+  namespace: default    # Since Config Data does not include "k8sKey", default namespace is used.
```
asoorm, line 170: what is `k8sKey`?

pkg/snapshot/snapshot.go
```diff
@@ -0,0 +1,389 @@
...
+
+	policiesResponse := &dashboard.PoliciesResponse{}
+
+	err = jsonEncoding.NewDecoder(res.Body).Decode(&policiesResponse)
+	if err != nil {
+		return nil, err
+	}
+
+	return policiesResponse.Policies, nil
+}
```
asoorm, line 275: This logic belongs in the Dashboard Client All method

api/model/api.go
buraksekili: Dashboard stores empty value for `http` APIProtocol. Once `http` APIs are exported from Dashboard, the corresponding YAML file has an empty APIProtocol value (`""`), which prevents applying that manifest directly due to validation failure. Users need to update empty values with `http`. That's why I added empty string, `""`, to the validation enum.

api/model/api.go
buraksekili: similar reasons as `APIProtocol` field. 
Once APIs are exported, all fields come with their nil values if they are not explicitly declared. So, although users do not use any GraphQL features, they cannot apply their manifests due to the empty value of `GraphQLExecutionMode` which is mandatory.

api/model/api.go
asoorm: Does it make sense to mutate the response from the Dashboard API to insert http in there? Perhaps we could fix it inside the defaulting webhook? 

Or with this change are we saying that this field is completely optional - and up to the dashboard / gateway API to decide what to do with it?

pkg/snapshot/README.md
```diff
@@ -0,0 +1,203 @@
...
+
+You can also dump all ApiDefinitions from Dashboard via `--all` flag.
+
+```bash
+TYK_MODE=pro TYK_URL=<TYK_URL> TYK_AUTH=<TYK_AUTH> TYK_ORG=<TYK_ORG> ./tyk-operator --snapshot output.yaml --all
+```
+
+#### Output CR
+
+`tyk-operator` CLI creates an output file specified via `--snapshot` flag. Each ApiDefinition CR metadata has a default name  `replace-me-{i}` where `{i}` increases by each ApiDefinition.
```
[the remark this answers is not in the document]
buraksekili: at the moment, the tool generates files with valid meta names to allow end-users to create manifests without further modifications. what do you think about this approach? should it be changed as you suggested (forcing the user to replace the value)?

cc: @caroltyk

pkg/snapshot/README.md
buraksekili: Thank you for the suggestion! It offers a better experience than it does presently.
I will update the PR accordingly

pkg/snapshot/README.md
buraksekili: sorry, it should be `k8sNamespace`. I will update it. Thanks for spotting it

pkg/snapshot/snapshot.go
buraksekili: @asoorm, `All` method of Policies returns paged results and It seems like the current client does not support fetching results page by page - or I am missing something.

As far as I understand from the policies API, I need to give `?p=-2` query parameter to fetch all policies from the dashboard. That's why I added a separate function with a specific query parameter. 

Should we update our client in a separate PR to fetch all policies without paging?

pkg/snapshot/README.md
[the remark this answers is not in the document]
asoorm: > allow end-users to create manifests without further modifications

I don't like this approach it's flaky as it's just an array. You can't get any consistency if APIs are created, updated, deleted in the source environment. that makes the feature effectively impossible to use in a real-life scenario.

If they want to create valid names - they should use the configData field in the API Definition to instruct the Operator what name to put in the definition. Otherwise, it should fail - and they should name the CRD manually.

pkg/snapshot/snapshot.go
asoorm: I would suggest that the current client is a bug - and we should edit the current client to return ALL policies using `?p=-2`

api/model/api.go
buraksekili: >Or with this change are we saying that this field is completely optional - and up to the dashboard / gateway API to decide what to do with it?

Actually, this is what I was trying to achieve - letting dashboard/gw decide on what to do with empty "protocol" since providing empty protocol does not raise any errors on the dashboard/gw side.

As you mentioned, we can update empty APIProtocol values in defaulting webhooks but I believe it might cause a problem since the dashboard has a mechanism to set it to a default value. It defaults `protocol` to your gw configuration if `protocol` is empty.

If you see this change as breaking, I can update snapshot tool logic to update empty APIProtocols as `http`.

asoorm: I think it's a good decision. Thanks.

pkg/snapshot/README.md
```diff
@@ -42,13 +42,18 @@ go build
 ## Preparation
 1. Use API Category to group the APIs you want to export
 
-By default, snapshot tool will export all APIs from Tyk dashboard with category `operator`.  You can also categorize your APIs by teams, environments, or any way you want. You can specify which API category to be exported in later steps.
+By default, snapshot tool will export all APIs from Tyk dashboard with category `operator`.  
+You can also categorize your APIs by teams, environments, or any way you want. 
+You can specify which API category to be exported in later steps.
```
asoorm, line 47: Why by default will operator only pull categorised `operator`? Default should be ALL apis, and perhaps optionally filter by category?

buraksekili: >Default should be ALL apis, and perhaps optionally filter by category?

Yes, exactly! Updated accordingly

pkg/snapshot/snapshot.go
buraksekili: Created an issue on GH: #503 
I will raise a PR asap.

pkg/snapshot/README.md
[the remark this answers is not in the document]
buraksekili: Updated as you suggested 👍 
Now, the default name is `REPLACE_ME_{i}` instead of `replace-me-{i}` so users need to update it either manually or using ConfigData.

pkg/snapshot/snapshot.go
```diff
@@ -0,0 +1,369 @@
...
+
+// parseConfigData parses given ApiDefinitionSpec's ConfigData field. It checks existence of NameKey and NamespaceKey
+// keys in the ConfigData map. Returns their values if keys exist. Otherwise, returns default values for name and namespace.
+func parseConfigData(apiDefSpec *model.APIDefinitionSpec, defName string) (name, namespace string) {
+	if apiDefSpec.ConfigData == nil {
+		return defName, DefaultNs
+	}
+
+	// Parse name
+	val, ok := apiDefSpec.ConfigData.Object[NameKey]
```
komalsukhani, line 265: Instead of doing same thing twice for NameKey and NamespaceKey, can you store them in array and use for loop?

pkg/snapshot/README.md
```diff
@@ -0,0 +1,225 @@
...
+
+[Prerequisites](#prerequisites) | [Installation](#installation) | [Preparation](#preparation) | [Usage](#usage) | [Limitations](#limitations)
+
+---
+
+## Prerequisites
+
+1. Access to `tyk-operator` repository.
+2. [go](https://go.dev/doc/install)
+3. Credentials to connect Tyk Dashboard. Please visit [Tyk Docs](https://tyk.io/docs/tyk-stack/tyk-operator/installing-tyk-operator/#tyk-self-managed-hybrid) for details.
```
komalsukhani, line 27:
```suggestion
3. Credentials to connect Tyk Dashboard/Gateway. Please visit [Tyk Docs](https://tyk.io/docs/tyk-stack/tyk-operator/installing-tyk-operator/#tyk-self-managed-hybrid) for details.
```
I believe it should work for both Pro and CE mode

pkg/snapshot/README.md
```diff
@@ -0,0 +1,225 @@
...
+
+## Installation
+
+1. Clone Tyk Operator repository
+```bash
+git clone https://github.com/TykTechnologies/tyk-operator.git
+cd tyk-operator/
+git checkout feat/importer-cli
```
komalsukhani, line 35: Once we merge this in master, this step won't be required.

pkg/snapshot/README.md
buraksekili: haven't tested for CE yet. I will test it and if it works, I will update the documentation accordingly 👍

pkg/snapshot/snapshot.go
buraksekili: Can you please elaborate on it? I couldn't fully understand. 
how can we understand that array elements belong to Name or Namespace?

asoorm:
Is there a reason thepkg/snapshot/README.md is not inside the docs directory?
It would also be great to reference this file from within the homepage so that it is easy to discover and navigate to.

controllers/securitypolicy_controller.go
```diff
@@ -227,11 +227,25 @@ func (r *SecurityPolicyReconciler) create(ctx context.Context, policy *tykv1.Sec
 		return err
 	}
 
-	err = klient.Universal.Portal().Policy().Create(ctx, spec)
-	if err != nil {
-		r.Log.Error(err, "Failed to create policy")
+	// Check if policy exists. During migration, policy exists on the Dashboard but not in the k8s environment. Therefore,
```
komalsukhani, line 230: Is it possible to call update function in that case?

pkg/snapshot/snapshot.go
komalsukhani: Since key name is fixed you can figure out based on that.

Or you can create a function so that we don't duplicating  the code.

pkg/snapshot/README.md
```diff
@@ -0,0 +1,225 @@
...
+so that you do not need to manually update the output file afterwards.
+
+![config-data](./img/config-data.png)
+
+## Usage
+```bash
+tyk-operator exports APIs and Security Policies from your Tyk Dashboard to Custom Resource that can be used with Tyk Operator
+
+Export API Definitions:
+   --snapshot <output_file>
```
komalsukhani, line 66:
```suggestion
  --snapshot <output_file>
```

To match indentation of other flags

komalsukhani: Also should we rename this to apidef or something to match with policy option?

pkg/snapshot/README.md
komalsukhani: @buraksekili Need to update screenshot. It is still referring to `k8sNs`.

pkg/snapshot/README.md
komalsukhani: If it doesn't work, we should mention that in doc.

pkg/snapshot/README.md
buraksekili: Resolved in bc5fd2fafe316d8209ba62045c61df99fd66e523

pkg/snapshot/snapshot.go
buraksekili: @komalsukhani added new function in bc5fd2fafe316d8209ba62045c61df99fd66e523 as you suggested. can you please check if it is okay for you?

pkg/snapshot/README.md
buraksekili: `--snapshot` flag looks good to me. do you have any suggestions?

pkg/snapshot/snapshot.go
komalsukhani: Looks great! Thank you.

pkg/snapshot/README.md
komalsukhani: It's clear from `--policy`that it will store policies in the file. But with `--snapshot` it's not clear that it will store APIDefinition in that file.
@asoorm @caroltyk @andrei-tyk What do you think?

pkg/snapshot/README.md
buraksekili: It works with CE (for only ApiDefinitions). Updated documentation 👍

komalsukhani:
@buraksekili Looks good to me. You will also need to update screenshot because it is still referring to `k8sNs` field of configData

buraksekili:
Updated @komalsukhani 👍

pkg/snapshot/README.md
caroltyk: I think --apidef would be clearer about what the outputs are (like --policy)

caroltyk:
Hi @buraksekili 

Thanks for the updates.

About the k8s metadata names - rather than output files with invalid metadata names in it, how about we make it mandatory to have "k8sName" configured in config data? Reasons for doing this:
1) It make the export process 'repeatable' without further manual changes required by user on output file. 
2) The security policies exported will also be usable with the correct reference of API names in access rights array.

Propose logic:
- Check if APIs have required config data (k8sName). If not, output a WARNING message and not include that API in output.

cc @asoorm @komalsukhani

asoorm:
> how about we make it mandatory to have "k8sName" configured in config data?

I love this - it's much cleaner and simpler - and makes it explicit which apis to export.

> * When exporting API, check if the API have required config data (k8sName). If not, output a WARNING message and not include that API in output.
> * When exporting policy, check If a policy reference to an API without k8sName config, output a WARNING message and not include that policy in output.

Need some clearer picture here why / how to warn the user - e.g. if `k8sName` doesn't exist how can we issue a warning when the output is a yaml? Perhaps it should just fail with non 0 exit code if they try to export an API without a k8sName?

pkg/snapshot/README.md
buraksekili: Updated accordingly in [916ff3b](https://github.com/TykTechnologies/tyk-operator/pull/481/commits/916ff3b7bcc2bdcdb1f5c52a218fd561997d9876). @caroltyk @komalsukhani

buraksekili:
> Need some clearer picture here why / how to warn the user - e.g. if `k8sName` doesn't exist how can we issue a warning when the output is a yaml? Perhaps it should just fail with non 0 exit code if they try to export an API without a k8sName?

@asoorm, what I had in mind is that if no valid config data is provided, let's not generate the output file and instead print a message to the console to inform users about invalid APIs. What do you think?

pkg/snapshot/snapshot.go
```diff
@@ -0,0 +1,335 @@
...
+	})
+
+	exportApiDefs := func(i int, w io.Writer, v *model.APIDefinitionSpec) error {
+		// Parse Config Data of the ApiDefinition created on Dashboard.
+		name, ns, err := parseConfigData(v, fmt.Sprintf("%s_%d", DefaultName, i))
+		if err != nil {
+			// Instead of terminating the function with error, just do not export ApiDefinitions with invalid ConfigData
+			// fields, instead inform users with the following log message.
+			fmt.Printf("WARNING: failed to parse API %v due to malformed ConfigData, err: %v\n", v.APIID, err)
+			return nil
```
komalsukhani, line 85: Shouldn't it return error here?

buraksekili: While exporting ApiDefinition, writing ApiDefinition to a file, if ApiDefinitoin has no valid ConfigData, we do not write that ApiDefinition into the output file; instead, we skip writing its content.

Assume that we have the following ApiDefinitions;
```
ApiDefinition - A (with valid Config Data)
ApiDefinition - B (with invalid Config Data)
ApiDefinition - c (with valid Config Data)
```

The output file will contain ApiDeifnition A and ApiDefinition C.

komalsukhani: All good on that end. But while processing policy group wise, we skip APIDefinitions if it doesn't have valid configData but according to Carol's comment here 

> When exporting policy, check If a policy reference to an API without k8sName config, output a WARNING message and not include that policy in output.

We should skip processing of that policy all together. But since this function is not returning any error, it will just skip processing of APIs with invalid configData.

buraksekili: thank you for spotting this 🙏 i will update the PR accordingly

buraksekili: Updated in e4e7fdfcb5b9a9f93839842e7589e68a57951cfe.

Can you please re-review?

pkg/snapshot/README.md
```diff
@@ -0,0 +1,233 @@
...
+tyk-operator exports APIs and Security Policies from your Tyk installation to Custom 
+Resource that can be used with Tyk Operator
+
+Export API Definitions:
+  --apidef <output_file>
+    	By passing an export flag, we are telling the Operator to connect to a Tyk
+    	 installation in order to pull a snapshot of ApiDefinitions from that 
+    	 environment and output as CR
+
+  --group
```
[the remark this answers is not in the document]
buraksekili: @komalsukhani I've updated the docs. Can you please re-check again?

andrei-tyk (approved):

PR #487, opened 2022-08-09
fix reference to loop iterator defect

### LINKED ISSUE #486: List method in Dashboard Client uses reference to loop iterator variable
## Expected Behavior

If I have 3 APIs loaded into Tyk, then the List Method in the API Client should return those 3 APIs.

## Current Behavior

When calling the List function - the same 3 APIs are returned.

https://github.com/TykTechnologies/tyk-operator/blob/master/pkg/client/dashboard/api.go#L118-L120

## Possible Solution

https://github.com/golang/go/wiki/CommonMistakes#using-reference-to-loop-iterator-variable

## Steps to Reproduce

Call the List method

asoorm:
Fixes #486

### DIFF
--- a/pkg/client/dashboard/api.go
+++ b/pkg/client/dashboard/api.go
@@ -116,6 +116,7 @@ func (Api) List(
 	var a model.APIDefinitionSpecList
 
 	for _, v := range o.Apis {
+		v := v
 		a.Apis = append(a.Apis, &v.ApiDefinition)
 	}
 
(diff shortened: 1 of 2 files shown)


andrei-tyk (approved):

buger:
Can you explain this fix?

asoorm:
Sure https://github.com/golang/go/wiki/CommonMistakes#using-reference-to-loop-iterator-variable

It's also detailed in the linked issue #486

PR #489, opened 2022-08-11
New upstream samples using cert-manager issued certs and secrets

rewsmith:
## Description
New upstream tls sample which makes use of a cert-manager issued certificate and k8s secret

## Motivation and Context
Customer request

Followed the instructions in the sample and checked that both a certificate and an API referencing that certificate for upstream tls exist in Tyk Dashboard

## Types of changes

## Checklist

  - [ ] `gofmt -s -w .`
  - [ ] `go vet ./...`
  - [ ] `golangci-lint run`

### DIFF
--- a/config/samples/httpbin_upstream_certmanager_cert.yaml
+++ b/config/samples/httpbin_upstream_certmanager_cert.yaml
@@ -1,0 +1,36 @@
+##### How to use upstream certificate refs:
+
+#
+# First apply this manifest using the command
+# "kubectl apply -f config/samples/httpbin_upstream_certmanager_cert.yaml -n tyk"
+#
+# The operator will try to create the ApiDefinition and will succeed but will log an error that a certificate is missing
+# in the cluster for an upstream
+
+# Apply the cert-manager issued TLS certificate using the following command: (we already have an example one in our repo)
+# "kubectl apply -f config/samples/tls/upstream/upstream-tls-cert-manager.yaml -n tyk"
+
+# NOTE: the upstream_certificate_refs can hold references to multiple secrets which are used for the domain
+# mentioned in the key (currently "*" is used as a wildcard for all the domains)
+
+apiVersion: tyk.tyk.io/v1alpha1
+kind: ApiDefinition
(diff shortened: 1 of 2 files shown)


config/samples/httpbin_upstream_certmanager_cert.yaml
```diff
@@ -0,0 +1,36 @@
+##### How to use upstream certificate refs:
+
+#
+# First apply this manifest using the command
+# "kubectl apply -f config/samples/httpbin_upstream_certmanager_cert.yaml -n tyk"
+#
+# The operator will try to create the ApiDefinition and will succeed but will log an error that a certificate is missing
```
buraksekili, line 7: Is there any reason why we should first create this ApiDefinition instead of the certificate? If not, maybe we can say something like "Operator can work even if the certificate is not in the cluster."

buraksekili (approved):
LGTM, thank you @rewsmith. I just have one remark. Can you please check it? If it is not crucial, we can merge it 👍

config/samples/httpbin_upstream_certmanager_cert.yaml
rewsmith: No reason - it will work just fine if the secret is created first.  Will amend and add a statement as you suggest

rewsmith: Actually I will change the order to create the certificate first, but also mention that it can work the other way around

komalsukhani:
I was thinking if we can place Certificate and API Def in same file under config/samples/tls
So we will have tls examples at same place.

komalsukhani:
Actually your idea of placing it tls/uptream sounds much better.

config/samples/httpbin_upstream_certmanager_cert.yaml
rewsmith: All done.  Ready to review again

buraksekili (approved):

buraksekili:
Thanks @rewsmith

PR #491, opened 2022-08-16
[TT-6314] Add custom Go auth plugin support

sedkis:
Adds support for go-auth plugins in APIDefinition CRD

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 ## [Unreleased](https://github.com/TykTechnologies/tyk-operator/tree/HEAD)
 [Full Changelog](https://github.com/TykTechnologies/tyk-operator/compare/v0.10.0...HEAD)
 
+** Custom Go Auth Plugin **
+- Added support for Go auth custom plugins
+
 ## [v0.10.0](https://github.com/TykTechnologies/tyk-operator/tree/v0.10.0)
 [Full Changelog](https://github.com/TykTechnologies/tyk-operator/compare/v0.9.0...v0.10.0)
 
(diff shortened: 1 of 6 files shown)


sedkis:
1.  Using [this go plugin example](https://gist.github.com/sedkis/0b4a3222ed57cc1a0b55d4c53b0ba10a#file-goplugin-go)

Works with the API Def YAML in the PR.  Has a "PRE" hook and a "AUTH" hook:

## test case 1, With Auth

```
curl localhost:8089/httpbin/get -H "Authorization: abc"
{
  "args": {},
  "headers": {
    "Accept": "*/*",
    "Accept-Encoding": "gzip",
    "Authorization": "abc",
    "Foo": "Bar",
    "Host": "httpbin.org",
    "User-Agent": "curl/7.77.0",
    "X-Amzn-Trace-Id": "Root=1-62fb7de5-5146869731e1b6c33873941a"
  },
  "origin": "127.0.0.1, 213.203.151.163",
  "url": "http://httpbin.org/get"
}
```

See the 200 above

## Without auth, get 403:
```
curl localhost:8089/httpbin/get
{"reason":"Access forbidden"}%
```

andrei-tyk:
Thank you for your work sedky and sorry for answering at such a late time and hour.

The logic and examples look fine but I feel that's just because I have already read https://tyk.io/docs/plugins/supported-languages/golang/

Given that someone would look for the example AFTER reading the docs I think it's safe to say that we do not need to leave a link in the example so therefore I'll approve your pull req. 

Thank you again for your help!

andrei-tyk (approved):

PR #502, opened 2022-09-02
Add unit tests using fake client

komalsukhani:
## Description

Added unit tests using [fake client](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client/fake)

## Related Issue

[TT-5177](https://tyktech.atlassian.net/browse/TT-5177)

## Motivation and Context

Improve unit test coverage of the project

Added unit tests 

## Types of changes

## Checklist

  - [ ] `make manifests`
  - [ ] `make helm`
  - [x] `gofmt -s -w .`
  - [x] `go vet ./...`
  - [x] `golangci-lint run`


controllers/ingress_controller_test.go
```diff
@@ -1,6 +1,18 @@
...
+	"github.com/matryer/is"
+	v1 "k8s.io/api/networking/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
 
 func TestShortHash(t *testing.T) {
 	if expect, got := "c09b5928e", shortHash("foo.com"+"/httpbin22222"); got != expect {
```
buraksekili, line 18: I know that this test case does not belong to the PR, but we can add more "unhappy" cases here. For example, what happens if domain or host names are empty?

controllers/operatorcontext_controller_test.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+func TestOperatorContextDelete(t *testing.T) {
+	is := is.New(t)
+	dummyName := "dummy"
+
+	tests := map[string]struct {
+		Name        string
+		OpCtxStatus *v1alpha1.OperatorContextStatus
+		ReturnError bool
```
buraksekili, line 69: Considering this is the initial work for unit tests, it looks good to me. But what do you think about checking for errors in `error` type instead of `bool` in the future? So, we can understand that we obtained the desired error instead of any errors.

buraksekili (approved):
lgtm!

controllers/operatorcontext_controller_test.go
komalsukhani: Updated tests.

controllers/ingress_controller_test.go
komalsukhani: Yes. We should do that.

PR #520, opened 2022-10-13
[TT-6703] Added integration tests using Venom framework

komalsukhani:
## Description

Added integration tests using venom framework.

## Related Issue

[TT-6703](https://tyktech.atlassian.net/browse/TT-6703)

## Types of changes

## Checklist

  - [ ] `make manifests`
  - [ ] `make helm`
  - [ ] `gofmt -s -w .`
  - [ ] `go vet ./...`
  - [ ] `golangci-lint run`


venom-tests/api-json-validation.yml
```diff
@@ -0,0 +1,38 @@
+name:  Test Json Validation Plugin
```
buraksekili, line 1: I believe this test suite is duplicated. There is also `venom-tests/QA-960-api-json-validation.yml` that runs the same test.

venom-tests/lib/applyManifest.yml
```diff
@@ -0,0 +1,12 @@
+executor: applyManifests
+input:
+  filename: {}
+steps: 
+  - type: exec
+    script: |
+      kubectl apply -f '{{.input.filename}}'
+      sleep 20
```
buraksekili, line 8: it looks okay. 
I am not sure but can we use `kubectl wait` command if `sleep` is used to wait for the resource to be Ready?

andrei-tyk: Sounds like a good ideea if we don't want to always lose those 20 secs.

venom-tests/samples/httpbin_json_schema_validation copy.yaml
```diff
@@ -0,0 +1,81 @@
...
+                    type: string
+                    minLength: 2
+                  age:
+                    type: integer
+                    minimum: 1
+                required:
+                  - userName
+                type: object
+
+# k apply -f config/samples/httpbin_json_schema_validation.yaml
```
andrei-tyk, line 42: Is this commented code block worth keeping ?

venom-tests/samples/httpbin_blacklist.yaml
```diff
@@ -0,0 +1,34 @@
+apiVersion: tyk.tyk.io/v1alpha1
```
buraksekili, line 1: with your suggested modification on `config/samples/httpbin_blacklist.yaml` file, this file will be identical to [ `config/samples/httpbin_blacklist.yaml`](https://raw.githubusercontent.com/TykTechnologies/tyk-operator/6e36bcf3d213ffefaa6cc3074d8e324d66f50097/config/samples/httpbin_blacklist.yaml). Maybe, we can consider using one sample, instead of maintaining two copies of it; unless I am missing some technical details.

komalsukhani: I have added separate copies of all the manifests we are using so that we can modify them if needed for testing purpose.
Also will have manifests in same folder instead of jumping to different one.

venom-tests/lib/applyManifest.yml
komalsukhani: Great idea!

venom-tests/samples/httpbin_blacklist.yaml
buraksekili: 👌 makes sense. then, let's keep them

venom-tests/lib/applyManifest.yml
komalsukhani: @buraksekili I am not able to find a condition for `kubectl wait` to check if resource was created successfully.
Do you know how to do it?

buraksekili (approved):
thank you!

venom-tests/lib/applyManifest.yml
buraksekili: it seems like `kubectl wait` is not feasible to use with CRDs. Let's keep this part as it is

PR #540, opened 2022-11-30
TT 3695 - Fix Security Policy Migration

### LINKED ISSUE #204: [TT-3695] Validate/Document managing of policies created on the dashboard
Ensuring what is documented here https://github.com/TykTechnologies/tyk-operator/blob/master/docs/policies/migration.md  works properly and the doces are up to date

buraksekili:
## Description

The migration steps described in [this documentation](https://github.com/TykTechnologies/tyk-operator/blob/v0.12.0/docs/policies/migration.md) do not work as expected. Setting the `id` field in the CR to the `_id` field of the existing Policy has some issues.

## Related Issue

Resolves #204
https://tyktech.atlassian.net/browse/TT-3695

## Motivation and Context

Once you migrate the existing policy from Dashboard to k8s, k8s state should include details of the spec that exists on Dashboard, and from now on, the source of truth must become k8s. Operator must restore subsequent changes from Dashboard in the next reconciliations.

Added test cases to cover:
- Migration of an existing Security Policy from Dashboard to k8s
- Ensuring k8s is the source of truth.
    - Drifts (updates/deletes) from Dashboard should be restored based on the k8s state in the next reconciliation.
- Creating a simple SecurityPolicy should update ApiDefinition's status if that SecurityPolicy refers to it
    - Ensure that corresponding resources are created on Tyk
- Deleting this SecurityPolicy must update ApiDefinition status.
    - Ensure that the corresponding SecurityPolicy is deleted from Tyk as well

## Types of changes

## Checklist

- [x] My change requires a change to the documentation.
  - [ ] `make manifests`
  - [ ] `make helm`
  - [x] `gofmt -s -w .`
  - [x] `go vet ./...`
  - [x] `golangci-lint run`


controllers/securitypolicy_controller.go
```diff
@@ -219,53 +228,78 @@ func (r *SecurityPolicyReconciler) update(ctx context.Context, policy *tykv1.Sec
...
 
-			return err
+			return nil, err
 		}
+
+		policy.Spec.MID = spec.MID
 	} else {
+		spec.AccessRightsArray = policy.Spec.AccessRightsArray
+
+		spec, err = r.spec(ctx, spec)
```
komalsukhani, line 260: Why is it creating a copy of itself here?

integration/securitypolicy_test.go
```diff
@@ -0,0 +1,389 @@
...
+						copySpec.Name = "Updating Existing Policy"
+
+						err := klient.Universal.Portal().Policy().Update(reqCtx, copySpec)
+						eval.NoErr(err)
+
+						// Ensure that policy is updated accordingly on Tyk Side.
+						newCopySpec, err := klient.Universal.Portal().Policy().Get(reqCtx, policyOnK8s.Status.PolID)
+						eval.NoErr(err)
+						eval.True(newCopySpec != nil)
+						eval.Equal(newCopySpec.Name, copySpec.Name)
```
komalsukhani, line 197: Shouldn't the name still match to what it was in the k8s spec(Instead of `Updating Existing Policy` it should be `existing-spec`)?
Here you are checking if the name got updated with new value you set on tyk Dashboard.

komalsukhani: Ohh right. Sorry my bad

komalsukhani (approved):

komalsukhani:
@buraksekili Thank you for resolving this complicated bug! It looks good to me.

integration/securitypolicy_test.go
```diff
@@ -0,0 +1,400 @@
...
+
+	spec := v1alpha1.SecurityPolicySpec{
+		Name:   "existing-spec",
+		State:  "draft",
+		Rate:   34,
+		Active: true,
+		Tags:   []string{"testing"},
+	}
+
+	hasSameValues := func(k8sSpec, tykSpec *v1alpha1.SecurityPolicySpec, k8sStatusID string) bool {
```
komalsukhani, line 41: `k8sStatusID` is not used

integration/securitypolicy_test.go
```diff
@@ -0,0 +1,400 @@
...
+				err = wait.For(
+					conditions.New(c.Client().Resources()).ResourceMatch(&policyCR, func(object k8s.Object) bool {
+						policyOnK8s, ok := object.(*v1alpha1.SecurityPolicy)
+						eval.True(ok)
+						eval.True(len(policyOnK8s.Status.PolID) > 0)
+
+						policyOnTyk, err := klient.Universal.Portal().Policy().Get(reqCtx, policyOnK8s.Status.PolID)
+						eval.NoErr(err)
+
+						eval.True(policyOnK8s.Spec.MID == policyOnTyk.MID &&
```
komalsukhani, line 130: Can't we use `hasSameValues()` here too?

integration/securitypolicy_test.go
```diff
@@ -0,0 +1,400 @@
...
+
+				err = wait.For(
+					conditions.New(c.Client().Resources()).ResourceMatch(&policyCR, func(object k8s.Object) bool {
+						policyOnK8s, ok := object.(*v1alpha1.SecurityPolicy)
+						eval.True(ok)
+
+						newSpec, err := klient.Universal.Portal().Policy().Get(reqCtx, policyOnK8s.Status.PolID)
+						eval.NoErr(err)
+
+						eval.True(hasSameValues(&policyOnK8s.Spec, newSpec, policyOnK8s.Status.PolID))
```
komalsukhani, line 176: Can you also check if the new and old(before deleting it from Dashboard) policyID are same?

integration/securitypolicy_test.go
```diff
@@ -0,0 +1,400 @@
...
+			}).
+		Assess("Must be restored from Dashboard updates to K8s state on the next reconciliation",
+			func(ctx context.Context, t *testing.T, c *envconf.Config) context.Context {
+				// Assume that the user updated the Policy from Dashboard, which creates a drift between k8s state
+				// and Tyk. In order to establish consistency, Operator must update Security Policy based on the
+				// k8s state on next reconciliation - so that k8s remains as a source of truth.
+				eval := is.New(t)
+
+				err := wait.For(
+					conditions.New(c.Client().Resources()).ResourceMatch(&policyCR, func(object k8s.Object) bool {
```
komalsukhani, line 194: I think we don't need `ResourceMatch` here since we are not doing any operator on k8s resource. We are changing things on Tyk side. You can use `err = wait.For(func() (done bool, err error){}`

integration/securitypolicy_test.go
buraksekili: uh, thank you for spotting it. i totally missed that. updated accordingly in d17880a1cd921024b5b82813d7c62c2021ebb2cd

integration/securitypolicy_test.go
buraksekili: yes, we can. thanks for the suggestion. updated as you suggested in d17880a1cd921024b5b82813d7c62c2021ebb2cd

integration/securitypolicy_test.go
buraksekili: I compared previous & new IDs. Can you please re-check again (d17880a1cd921024b5b82813d7c62c2021ebb2cd) if it looks okay?

integration/securitypolicy_test.go
buraksekili: You are right, we only run `DeepCopy` on resource's Spec. So, we do not need `conditions`. Updated accordingly 👍 thanks!

integration/securitypolicy_test.go
komalsukhani: Ohh right.

komalsukhani (approved):

integration/securitypolicy_test.go
```diff
@@ -170,9 +170,13 @@ func TestSecurityPolicyMigration(t *testing.T) {
 
 						newSpec, err := klient.Universal.Portal().Policy().Get(reqCtx, policyOnK8s.Status.PolID)
 						eval.NoErr(err)
-						eval.True(newSpec.MID == previousPolicyID)
+						eval.True(hasSameValues(&policyOnK8s.Spec, newSpec, policyOnK8s.Status.PolID))
```
komalsukhani, line 173: You have added same check below too

buraksekili: the one below ensures that we can access the Policy by using the previous ID and it checks that the Policy that is accessed via the previous ID is the same as the one stored in Kubernetes.

PR #541, opened 2022-12-02
ignore 404 when deleting an API, fixes #469

### LINKED ISSUE #469: [TT-5482] Tyk Operator ignores "-f" during reconciliation infinite loops
As a developer, if I pass the `-f` flag to kubectl command to delete a resource, I expect it to delete the resource. However, the Tyk Operator hangs infinitely during the reconciliation loop.

## Recreate Steps
As a developer, if I (accidentally) delete an API/Policy in the Dashboard that was created through the Operator, and then I try to delete the 

IE, when I try to modify the resource through Operator, it hangs forever because the Dashboard resource does not exist.

## Expected Behavior
If I pass the `--force` command to the `kubectl delete`, the Operator should respect that.

## Possible Solution
Here, we should skip the reconciliation process if the `-f / --force` command is passed in from kubectl in addition to the current check that looks for a finalizer
https://github.com/TykTechnologies/tyk-operator/blob/master/controllers/apidefinition_controller.go#L482

patriziobruno:
Ignore not-found error when attempting to delete an API.

## Related Issue
#469

## Motivation and Context
`ApiDefinition` `CR` get stuck when API-definition has been deleted using the dashboard.

- Create an API-definition using `tyk-operator`
- Delete the API-definition using the dashboard
- Delete the API-definition using `tyk-operator`
- Verify that `tyk-operator` removed the api-definition `CR`

## Types of changes

## Checklist

  - [ ] `make manifests`
  - [ ] `make helm`
  - [ ] `gofmt -s -w .`
  - [ ] `go vet ./...`
  - [ ] `golangci-lint run`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ from Dashboard to Kubernetes environment. [#481](https://github.com/TykTechnolog
 - Fixed user email format used in integration tests since e2e tests were failing after Tyk v4.0.6 was released [#510](https://github.com/TykTechnologies/tyk-operator/pull/510)
 - Fixed bug in linking logic of SubGraph CR and ApiDefinition CR [#522](https://github.com/TykTechnologies/tyk-operator/pull/522).
 - Operator was panicking when invalid certificate was provided [#529](https://github.com/TykTechnologies/tyk-operator/pull/529)
+- Operator was failing to remove finalizers from ApiDefinition that was already deleted in Dashboard
 
 ## [v0.11.0](https://github.com/TykTechnologies/tyk-operator/tree/v0.11.0)
 [Full Changelog](https://github.com/TykTechnologies/tyk-operator/compare/v0.10.0...v0.11.0)
--- a/controllers/apidefinition_controller.go
+++ b/controllers/apidefinition_controller.go
@@ -542,7 +542,9 @@ func (r *ApiDefinitionReconciler) delete(ctx context.Context, desired *tykv1alph
 		r.Log.Info("deleting api")
 
 		_, err = klient.Universal.Api().Delete(ctx, desired.Status.ApiID)
-		if err != nil {
+		if err != nil && err.Error() == "Resource not found" {
+			r.Log.Error(err, "ignoring not existing api on delete", "api_id", desired.Status.ApiID)
+		} else if err != nil {
 			r.Log.Error(err, "unable to delete api", "api_id", desired.Status.ApiID)
 			return 0, err
 		}
--- a/controllers/apidefinition_controller.go
+++ b/controllers/apidefinition_controller.go
@@ -543,7 +543,7 @@ func (r *ApiDefinitionReconciler) delete(ctx context.Context, desired *tykv1alph
 
 		_, err = klient.Universal.Api().Delete(ctx, desired.Status.ApiID)
 		if err != nil && err.Error() == "Resource not found" {
-			r.Log.Error(err, "ignoring not existing api on delete", "api_id", desired.Status.ApiID)
+			r.Log.Error(err, "ignoring nonexistent api on delete", "api_id", desired.Status.ApiID)
 		} else if err != nil {
 			r.Log.Error(err, "unable to delete api", "api_id", desired.Status.ApiID)
 			return 0, err


controllers/apidefinition_controller.go
```diff
@@ -542,7 +542,9 @@ func (r *ApiDefinitionReconciler) delete(ctx context.Context, desired *tykv1alph
 		r.Log.Info("deleting api")
 
 		_, err = klient.Universal.Api().Delete(ctx, desired.Status.ApiID)
-		if err != nil {
+		if err != nil && err.Error() == "Resource not found" {
```
komalsukhani, line 545: Could you use this function to check if error is `Resource not found` https://github.com/TykTechnologies/tyk-operator/blob/c4b85072dce3a66eae6dce59e1d7eb20090f8a86/pkg/client/client.go#L38

komalsukhani:
Can you add tests for this scenario?

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ from Dashboard to Kubernetes environment. [#481](https://github.com/TykTechnolog
 - Fixed user email format used in integration tests since e2e tests were failing after Tyk v4.0.6 was released [#510](https://github.com/TykTechnologies/tyk-operator/pull/510)
 - Fixed bug in linking logic of SubGraph CR and ApiDefinition CR [#522](https://github.com/TykTechnologies/tyk-operator/pull/522).
 - Operator was panicking when invalid certificate was provided [#529](https://github.com/TykTechnologies/tyk-operator/pull/529)
+- Operator was failing to remove finalizers from ApiDefinition that was already deleted in Dashboard
```
komalsukhani, line 19: You need to make an entry under [Unreleased section](https://github.com/TykTechnologies/tyk-operator/blob/585124ee692701f3bd80301988be3d91508889b0/CHANGELOG.md#unreleased)

controllers/apidefinition_controller.go
patriziobruno: That was my first thought, but I wasn't sure about importing that package only for the error check. I'll go ahead and use `client.IsNotFound`

patriziobruno:
> Can you add tests for this scenario?

I could if I knew how to call the test dashboard-api to delete the API before tyk-operator can. Any pointers on how to call dashboard-api from within the BDD context?

bdd/godogs_test.go
```diff
@@ -253,6 +266,30 @@ func (s *store) iRequestEndpoint(path string) error {
 	)
 }
 
+func (s *store) iRequestDeleteDashEndpoint(path string) error {
+	return call(
+		http.MethodDelete,
+		createDashURL(path),
```
buraksekili, line 272: Can you please use `createURL` function?

bdd/godogs_test.go
```diff
@@ -60,8 +61,14 @@ var opts = &godog.Options{
 	StopOnFailure: true,
 	Format:        "pretty",
 	Tags:          "~@undone",
+	Strict:        true,
```
buraksekili, line 64: At the moment, it might be better not to use Strict mode since we are not highly depending on bdd tests.

buraksekili (commented):
this looks great to me! thank you @patriziobruno !

I would like to explain something about current `bdd` tests. At the moment, it is not able to use Dashboard API. It only uses Gateway API. That's why adding a step about dashboard API won't help in bdd tests. So, can you please update your tests; so that it only uses Gateway API instead of Dashboard API? 
Moreover, We decided to move with [Venom tests](https://github.com/TykTechnologies/tyk-operator/tree/master/venom-tests) instead of bdd tests. In the future, our plan is to migrate bdd tests to Venom gradually. Sorry for the confusion.

bdd/godogs_test.go
buraksekili: >if I call DELETE http://localhost:8080/api/apis/httpbin-delete-test will it delete the api-definition?

you can call `DELETE http://localhost:8080/tyk/apis/${API_ID}` - instead of `api/apis`, GW API uses `tyk/apis`. you can find more details [here](https://tyk.io/docs/tyk-gateway-api/) if you wish.

again, thank you for your contribution

buraksekili (approved):
LGTM unless anyone else has any opinions! Once QA finishes testing this PR, we can merge it for the next release. thank you for your great effort @patriziobruno !

caroltyk:
Thanks for the contribution @patriziobruno ! The change is now merged and will be included in the next release targeted at early Jan.

PR #547, opened 2022-12-15
[TT-5089] Add new field in security policy status

### LINKED ISSUE #431: [TT-5089] - Attempting to remove an ApiDefinition fails if previously associated to a SecurityPolicy
Tyk Hybrid Gateway Version: 3.2.1
Tyk Cloud Gateway Dashboard Version: 3.2.0
Tyk Operator Version: 0.8.2

Full logs attached [logs.zip](https://github.com/TykTechnologies/tyk-operator/files/8453758/logs.zip)

Reproduction steps:

1. Create an ApiDefinition and SecurityPolicy that are associated

```yaml
kind: ApiDefinition
metadata:
  name: httpbin
spec:
  name: httpbin protected
  protocol: http
  active: true
  proxy:
    target_url: http://httpbin.org
    listen_path: /httpbin
    strip_listen_path: true
  use_standard_auth: true
  auth_configs:
    authToken:
      auth_header_name: Authorization
```

```yaml
kind: SecurityPolicy
metadata:
  name: httpbin
spec:
  id: 62505570b6e263000139f7ea
  name: Httpbin Security Policy
  state: active
  active: true
  access_rights_array:
    - name: httpbin
      namespace: default
      versions:
        - Default
```
2.  Remove the `access_rights_array` from the SecurityPolicy.
3.  Attempt to remove the ApiDefinition, but this error is returned:
```bash
{"level":"info","ts":1649434399.083711,"logger":"controllers.ApiDefinition","msg":"checking linked security policies"}
{"level":"error","ts":1649434399.0837798,"logger":"controller-runtime.manager.controller.apidefinition","msg":"Reconciler error","reconciler group":"tyk.tyk.io","reconciler kind":"ApiDefinition","name":"httpbin","namespace":"default","error":"unable to delete api due to security policy dependency=default/httpbin","stacktrace":"sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.9.0/pkg/internal/controller/controller.go:253\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.9.0/pkg/internal/controller/controller.go:214"}
{"level":"info","ts":1649434401.644129,"logger":"controllers.ApiDefinition","msg":"Reconciling ApiDefinition instance","ApiDefinition":"default/httpbin"}
{"level":"info","ts":1649434401.6443067,"logger":"controllers.ApiDefinition","msg":"resource being deleted"}
``` 

It appears that removing the API from the `access_rights_array` does not properly remove the SecurityPolicy from the `linked_by_policies` in the ApiDefinition:

```yaml
  linked_by_policies:
    Name:       httpbin
    Namespace:  default
```

komalsukhani:
## Description

APIs  store link to security policies in their status field. It is used to prevent deletion of API if it is linked to policy.

When we remove an API from access rights of policy, link stored in API is not deleted. This happens because there is no way for the security policy controller to know which API was removed. It knows only about the current state.
To fix this issue, I have added `LinkedAPIs` field in the status of the policy. As the name says, it stores list of APIs it is linked to.
Now, security policy controller has information about the APIs that were linked before.

## Related Issue

[TT-5089](https://tyktech.atlassian.net/browse/TT-5089)
Fixes #431 

## Motivation and Context

We should be able to delete API which is not linked to the policy.

Added integration and unit tests

## Types of changes

## Checklist

  - [ ] `make manifests`
  - [ ] `make helm`
  - [x] `gofmt -s -w .`
  - [x] `go vet ./...`
  - [x] `golangci-lint run`


integration/apidefinition_test.go
```diff
@@ -77,30 +77,20 @@ func TestApiDefinitionJSONSchemaValidation(t *testing.T) {
 				apiDef.Spec.VersionData.Versions = map[string]model.VersionInfo{
 					defaultVersion: {Name: defaultVersion, UseExtendedPaths: true, ExtendedPaths: eps},
 				}
-			}, envConf)
+			})
 			eval.NoErr(err) // failed to create apiDefinition
 
+			apiDef := v1alpha1.ApiDefinition{
```
buraksekili, line 83: Sorry, I couldn't understand this part. Why we are creating an ApiDefinition on Tyk here?

integration/helper_test.go
```diff
@@ -84,17 +91,39 @@ func generateSubGraphCR(namespace string, mutateFn func(graph *v1alpha1.SubGraph
...
 
 	err := c.Resources(namespace).Create(ctx, apiDef)
+	if err != nil {
+		return nil, err
+	}
 
 	return apiDef, err
 }
 
+func waitForTykResourceCreation(envConf *envconf.Config, obj k8s.Object) error {
```
buraksekili, line 108: this looks great 💯

integration/apidefinition_test.go
komalsukhani: I need to pass that object to `waitForTykResourceCreation`

buraksekili: okay 👍  
instead of calling `waitForTykResourceCreation` here, shouldn't creating a resource in k8s create a resource in Tyk? because previously, it seems like we were not creating resources on Tyk manually.

komalsukhani: Sometimes it takes time to create resource on Tyk. Because of which successive operations fails.
That's why I have added `waitForTykResourceCreation` in situations we want to make sure API is created on Tyk before proceeding.

buraksekili: 👍 thanks for the explanation

integration/securitypolicy_test.go
```diff
@@ -0,0 +1,178 @@
...
+
+				if len(pol.Status.LinkedAPIs) == 2 {
+					return true
+				}
+
+				return false
+			}), wait.WithTimeout(defaultWaitTimeout), wait.WithInterval(defaultWaitInterval))
+			is.NoErr(err)
+
+			is.True(pol.Status.LinkedAPIs[0].Name == api2Name || pol.Status.LinkedAPIs[1].Name == api2Name)
```
buraksekili, line 123: here, we only check whether CR is updated or not. should we check that updating `access_rights_array` affects the Security Policy on Tyk? What do you think?

integration/securitypolicy_test.go
```diff
@@ -0,0 +1,178 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/e2e-framework/klient/k8s"
+	"sigs.k8s.io/e2e-framework/klient/wait"
+	"sigs.k8s.io/e2e-framework/klient/wait/conditions"
+	"sigs.k8s.io/e2e-framework/pkg/envconf"
+	"sigs.k8s.io/e2e-framework/pkg/features"
+)
+
+func TestSecurityPolicyStatusIsUpdated(t *testing.T) {
+	is := is.New(t)
```
buraksekili, line 19: Maybe we should rename `is` to `eval` or something else to prevent duplicated declarations (package name `is` and local variable `is` ).

controllers/securitypolicy_controller.go
```diff
@@ -248,46 +243,88 @@ func (r *SecurityPolicyReconciler) create(ctx context.Context, policy *tykv1.Sec
...
 	policy.Status.PolID = policy.Spec.MID
 
+	if policy.Spec.AccessRightsArray != nil && len(policy.Spec.AccessRightsArray) > 0 {
+		policy.Status.LinkedAPIs = make([]model.Target, 0)
+	} else {
+		policy.Status.LinkedAPIs = nil
+	}
+
+	for _, v := range policy.Spec.AccessRightsArray {
+		target := model.Target{Name: v.Name, Namespace: v.Namespace}
```
buraksekili, line 271: what if `v.Namespace` is an empty string? Should we assign it to `policy.Namespace` if empty?

integration/securitypolicy_test.go
komalsukhani: I didn't add that because I thought we should have a separate tests to check policies are getting updated on Tyk.
I think you added that in your tests if I remember correctly.

controllers/securitypolicy_controller.go
komalsukhani: I don't think that's good idea. If namespace is invalid, then it should not even call this function.
It should fail before.

buraksekili (approved):

PR #558, opened 2023-01-10
Add an ad-hoc Tyk API call to verify non-existent Policies in delete method

buraksekili:
## Description
The CI fails due to failures that occurred in security policy deletion. As of today, the Policy API on GW does not return `404` while deleting non-existent Policies. Therefore, Operator tries to reconcile and cannot understand the existence of Policies. Until a fix is released on GW, this PR adds an additional step to verify the existence of Policy objects on Tyk level.
## Related Issue

https://github.com/TykTechnologies/tyk-operator/actions/runs/3873378749/jobs/6603330308#step:8:29

## Types of changes

## Checklist

  - [ ] `make manifests`
  - [ ] `make helm`
  - [ ] `gofmt -s -w .`
  - [ ] `go vet ./...`
  - [ ] `golangci-lint run`

### DIFF
--- a/controllers/securitypolicy_controller.go
+++ b/controllers/securitypolicy_controller.go
@@ -195,6 +195,15 @@ func (r *SecurityPolicyReconciler) delete(ctx context.Context, policy *tykv1.Sec
 			return nil
 		}
 
+		// At the moment, the Policy API returns 500 instead of 404 in cases of deleting non-existent Policies.
+		// So that reconciliation fails since non-existence couldn't be understood by Operator. Until this
+		// issue is fixed on Gateway level, add an ad-hoc API call to verify that policy's existence on Tyk level.
+		// If the Policy does not exist, no need to reconcile again. Otherwise, return err and reconcile.
+		_, errTyk := klient.Universal.Portal().Policy().Get(ctx, policy.Status.PolID)
+		if opclient.IsNotFound(errTyk) {
+			return nil
+		}
+
 		r.Log.Error(err, "Failed to delete resource from Tyk")
 
 		return err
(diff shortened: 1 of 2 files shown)


controllers/securitypolicy_controller.go
```diff
@@ -195,6 +195,15 @@ func (r *SecurityPolicyReconciler) delete(ctx context.Context, policy *tykv1.Sec
 			return nil
 		}
 
+		// At the moment, the Policy API returns 500 instead of 404 in cases of deleting non-existent Policies.
+		// So that reconciliation fails since non-existence couldn't be understood by Operator. Until this
+		// issue is fixed on Gateway level, add an ad-hoc API call to verify that policy's existence on Tyk level.
+		// If the Policy does not exist, no need to reconcile again. Otherwise, return err and reconcile.
+		_, errTyk := klient.Universal.Portal().Policy().Get(ctx, policy.Status.PolID)
+		if opclient.IsNotFound(errTyk) {
+			return nil
```
komalsukhani, line 204: Instead of returning here, we should continue to perform operations defined below if.
```
err := r.updateStatusOfLinkedAPIs(ctx, policy, true)
	if err != nil {
		return err
	}

	err = klient.Universal.HotReload(ctx)
	if err != nil {
		r.Log.Error(err, "Failed to hot-reload Tyk after deleting a Policy",
			"Policy", client.ObjectKeyFromObject(policy),
		)

		return err
	}
```

buraksekili: I updated can you please re-review again? Now, the execution continues only if `errTyk` is `404`. The other response codes 200 or even 500 make operator continue working.

komalsukhani (approved):

PR #605, opened 2023-04-04
TT-8234 - fix overriding contextRef while updating policy spec

buraksekili:
## Types of changes

## Checklist

  - [ ] `make manifests`
  - [ ] `make helm`
  - [ ] `gofmt -s -w .`
  - [ ] `go vet ./...`
  - [ ] `golangci-lint run`


integration/helper_test.go
```diff
@@ -160,12 +165,14 @@ func createTestPolicy(ctx context.Context, c *envconf.Config, namespace string,
...
 
-	policy.Name = testSecurityPolicy
+	policy.Name = testSecurityPolicy + fmt.Sprintf("%d", rand2.Int())
 	policy.Namespace = namespace
 	policy.Spec = v1alpha1.SecurityPolicySpec{
-		Name:   testSecurityPolicy,
-		Active: true,
-		State:  "active",
+		SecurityPolicySpec: model.SecurityPolicySpec{
+			Name:   testSecurityPolicy + testSecurityPolicy + uuid.New().String()[:8],
```
komalsukhani, line 172: Why are you using just first 8 characters of string here?

integration/securitypolicy_test.go
```diff
@@ -804,3 +804,200 @@ func TestSecurityPolicyForGraphQL(t *testing.T) {
...
+		Assess("Break the link between resources by removing reference",
+			func(ctx context.Context, t *testing.T, c *envconf.Config) context.Context {
+				testNs, ok := ctx.Value(ctxNSKey).(string)
+				eval.True(ok)
+
+				policy.Spec.Context = nil
+				err := c.Client().Resources(testNs).Update(ctx, policy)
+				eval.NoErr(err)
+
+				err = wait.For(conditions.New(c.Client().Resources()).
```
komalsukhani, line 968: Maybe we can skip this check. Since we are just setting field to nil and don't perform any operation in the controller.

buraksekili: yes, you are right. i'll update the test case accordingly

integration/helper_test.go
buraksekili: no specific reason. i just used 8 characters to prevent possible naming overflows etc.

komalsukhani: If we use just 8 characters, not sure it will be always a unique string.

integration/securitypolicy_test.go
buraksekili: @komalsukhani, on second thought, maybe we can keep this test case here. i remember my goal here to understand that setting `contextRef` to nil won't change in later steps. I mean, we have a logic where we handle deleting/adding context references to resources. so, I just wanted to ensure that after deleting a `contextRef`, the controller won't add the same contextRef to resources in subsequent reconciliations. what do you think

komalsukhani: Makes sense

integration/helper_test.go
buraksekili: updated as you suggested 👍 thanks!

komalsukhani (approved):

PR #607, opened 2023-04-10
[TT-8499] Fix panic caused while uploading certificate

komalsukhani:
## Description

- Refactored secret-cert-controller code
- Fixed panic which was caused by a assignment of value to nil slice. Fixed in commit [032789e](https://github.com/TykTechnologies/tyk-operator/pull/607/commits/032789ed67eb326c63153850b60f48c25cdd4a4e)

## Related Issue

[TT-8499](https://tyktech.atlassian.net/browse/TT-8499)

## Motivation and Context

Operator was panicking while setting `Certificates` field of APIDefinition when talking to OSS gateway.

- Added integration tests 
- Verified the issue is not reproducible.

## Types of changes

## Checklist

  - [ ] `make manifests`
  - [ ] `make helm`
  - [x] `gofmt -s -w .`
  - [x] `go vet ./...`
  - [x] `golangci-lint run`

[TT-8499]: https://tyktech.atlassian.net/browse/TT-8499?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


controllers/secret_cert_controller.go
```diff
@@ -168,33 +142,32 @@ func (r *SecretCertReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
-					// Requeue the Pod to try to reconciliate again.
+				// The Pod has been updated/deleted since we read it.
+				// Requeue the Pod to try to reconciliate again.
+				if apierrors.IsConflict(err) || apierrors.IsNotFound(err) {
 					return ctrl.Result{Requeue: true}, nil
 				}
 
 				if err != nil {
-					log.Error(err, "unable to update ApiDef")
+					log.Error(err, "Unable to update API Definition with cert id", "apiID", apiDefList.Items[idx].Spec.APIID)
```
buraksekili, line 152: Why isn't the error returned here? Correct me if I am wrong but it seems like we may need to trigger reconciliation again to update the corresponding ApiDefinition correctly. Also, if we return `err` here, `isCertAlreadyUploaded` will become `false` in the next reconciliation. so, the controller tries to upload an existing key. Maybe we should use Tyk APIs to verify the existence of keys - ofc, it comes with extra call costs.

komalsukhani: Opps sorry. Didn't realize that error is not returned. Will fix it.

> Also, if we return err here, isCertAlreadyUploaded will become false in the next reconciliation. so, the controller tries to upload an existing key. Maybe we should use Tyk APIs to verify the existence of keys - ofc, it comes with extra call costs.
It's fine if it re-updates certificate in next reconcile call. I just wanted to avoid unnecessary upload calls when certificate is already updated.

buraksekili (approved):

PR #610, opened 2023-04-18
Add ability to manage oidc auth

zalbiraw:
Add the ability to authenticate via OIDC

The OpenIDOptions  which holds the configuration for the OIDC auth has sub config called ClientIDs  of type map[string]string
The map consists of base64 of the OIDC client-id and policy ID as a key value pair.

The obscure issue I ran into is that the client-id is standard base64 and values like `oidc`, `cdio`, etc.. are a valid standard base64 so the operator can never determine if the client-id with value such as `oidc` should be transformed into base64 or not.

We can either document this or we tell the customer to provide the base64 of the client rather than the actual client id.

```
  name: oidc
  openid_options:
    providers:
    - client_ids:
        b2lkYy1jbGllbnQ=: dHlrL29pZGM
      issuer: https://keycloak-service.tyk.svc:7001/realms/oidc
    segregate_by_client: false
  protocol: http
  proxy:
```

Decided to ask the customer to provide a base64 value for the client-id

## Types of changes

## Checklist

  - [ ] `make manifests`
  - [ ] `make helm`
  - [ ] `gofmt -s -w .`
  - [ ] `go vet ./...`
  - [ ] `golangci-lint run`


zalbiraw:
Looking to deprecate OIDC support no need to include as JWT supports the same use case

PR #625, opened 2023-05-15
[TT-8822] Improve hash comparision in Reconciliation

buraksekili:
## Description

This PR improves the current hash comparison used to determine whether to send PUT requests or not, in order to update a resource.

It adds 2 status fields to ApiDefinition and SecurityPolicy CRDs
- `latestTykSpecHash` which stores the hash of lastly fetched resource from Tyk
- `latestCRDSpecHash` which stores the hash of lastly reconciled CR.

## Types of changes

## Checklist

  - [ ] `make manifests`
  - [ ] `make helm`
  - [ ] `gofmt -s -w .`
  - [ ] `go vet ./...`
  - [ ] `golangci-lint run`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased](https://github.com/TykTechnologies/tyk-operator/tree/HEAD)
 
+
+**Fixed**
+- Fixed panic of snapshot tool
+
 **Changed**
 - Changed optional fields of type string and bool to pointers for APIDefinition and Security Policy Custom Resources
 
--- a/go.mod
+++ b/go.mod
@@ -50,7 +50,7 @@ require (
 	github.com/jensneuse/abstractlogger v0.0.4 // indirect
 	github.com/jensneuse/byte-template v0.0.0-20200214152254-4f3cf06e5c68 // indirect
 	github.com/jensneuse/pipeline v0.0.0-20200117120358-9fb4de085cd6 // indirect
-	github.com/json-iterator/go v1.1.11 // indirect
+	github.com/json-iterator/go v1.1.12 // indirect
 	github.com/klauspost/compress v1.13.1 // indirect
 	github.com/matttproud/golang_protobuf_extensions v1.0.2-0.20181231171920-c182affec369 // indirect
 	github.com/mitchellh/copystructure v1.0.0 // indirect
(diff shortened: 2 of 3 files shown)


controllers/common.go
```diff
@@ -19,50 +19,31 @@ import (
 
 var hashOptions = hashstructure.HashOptions{ZeroNil: true}
 
-// isSameApiDefinition calculates the hash of two ApiDefinitionSpec structs and returns true
-// if hashes are equal. It is used to prevent calling extra Update requests to Tyk Gateway
-// if there is no changes on resources.
-func isSameApiDefinition(apiDef1, apiDef2 *model.APIDefinitionSpec) bool {
-	api1Hash, err1 := hashstructure.Hash(apiDef1, hashstructure.FormatV2, &hashOptions)
+// calculateHashes calculates hashes of given two interfaces. Returns empty string for hash
```
andrei-tyk, line 22: This function could be refactored to a single function that takes in only 1 argument and returns it's hash in the required format. Then it could be called twice where needed.

controllers/common.go
```diff
@@ -19,50 +19,31 @@ import (
...
-
-	pol1Hash, err1 := hashstructure.Hash(policy1, hashstructure.FormatV2, &myHashOptions)
-	if err1 == nil {
-		pol2Hash, err2 := hashstructure.Hash(policy2, hashstructure.FormatV2, &myHashOptions)
-		if err2 == nil {
-			return pol1Hash == pol2Hash
-		}
+// isSame compares given hash string with the hash of given interface and return true
+// if these values are same; otherwise, returns false.
+func isSame(latestHash string, i interface{}) bool {
```
andrei-tyk, line 40: If we break the "calculateHashes" function in a single parameter-single variable as suggested, this function can be removed all together and you cna use string == calculateHash(x)

integration/apidefinition_test.go
```diff
@@ -32,6 +32,11 @@ import (
 	"sigs.k8s.io/e2e-framework/pkg/features"
 )
 
+var (
```
andrei-tyk, line 35: these should be const

integration/securitypolicy_test.go
```diff
@@ -27,6 +27,11 @@ import (
 	"sigs.k8s.io/e2e-framework/pkg/features"
 )
 
+var (
```
andrei-tyk, line 30: these should be const

buraksekili: Yes but the returned value from `fmt.Errorf` or `errors.New` are not feasible to assign constant values.

integration/apidefinition_test.go
buraksekili: the returned value from `fmt.Errorf` or `errors.New` are not feasible to assign constant values.

andrei-tyk: you can return them as "constantVar + %v"

controllers/common.go
buraksekili: Thanks! Implemented in [0365448](https://github.com/TykTechnologies/tyk-operator/pull/625/commits/0365448d7f6ea2d6ea9cf16a4aa4bc1a16a72ec7)

controllers/common.go
buraksekili: Thanks! Implemented in [0365448](https://github.com/TykTechnologies/tyk-operator/pull/625/commits/0365448d7f6ea2d6ea9cf16a4aa4bc1a16a72ec7)

integration/securitypolicy_test.go
buraksekili: Updated [cb14d3b](https://github.com/TykTechnologies/tyk-operator/pull/625/commits/cb14d3bd5a47e5a6d7ec7938be244525cbc4652a)

PR #635, opened 2023-06-13
apidefinition JWTDefaultPolicies configured using operator cannot be configured in Tyk

owenchenxy:
change security policy id from the base64 encoded string of its name to its MID;
update the related api definition's jwt default policies while reconciling security policy;


komalsukhani:
Thank you for your contribution @owenchenxy 
Unfortunately we won't be able to accept this PR as we would to fix issue in the Dashboard UI instead of changing the functionality of operator.

caroltyk:
Hi @owenchenxy , this issue has been fixed in Dashboard 5.0.7 and 5.2.2. Please update Tyk version and try again. Thanks!

PR #670, opened 2024-01-04
[TT-10633] Add GraphQLIntrospectionConfig struct to ApiDefinition CRD

buraksekili:
This PR reverts changes of graphql introspection PR #669.
## Related Issue

https://tyktech.atlassian.net/browse/TT-10633

## Types of changes

## Checklist

  - [ ] `make manifests`
  - [ ] `make helm`
  - [ ] `gofmt -s -w .`
  - [ ] `go vet ./...`
  - [ ] `golangci-lint run`


api/model/api.go
```diff
@@ -1163,6 +1163,13 @@ type GraphQLConfig struct {
 
 	// Supergraph holds the configuration for a GraphQL federation supergraph.
 	Supergraph GraphQLSupergraphConfig `json:"supergraph,omitempty"`
+
+	// Introspection holds the configuration for GraphQL Introspection
+	Introspection GraphQLIntrospectionConfig `json:"introspection,omitempty"`
+}
+
+type GraphQLIntrospectionConfig struct {
+	Disabled bool `json:"disabled,omitempty"`
```
komalsukhani, line 1172: Can we set default value to true?

komalsukhani (changes requested):

api/model/api.go
buraksekili: @komalsukhani  on second thought, I propose setting this value to false by default, as it aligns with users' expectations to see the result of GraphQL introspection by default. For those who prefer an alternative to disable introspection, a configuration option is available on Tyk side to disable it. Opting for a default of true might potentially lead to confusion. What are your thoughts on this approach?

komalsukhani: Won't it cause problem as API Definition will not match with CRD?

buraksekili: I do not think that as APIDefinition on Tyk is only compared with the hash of ApiDefinition on Tyk that is obtained after creating CRD.

komalsukhani: okay. We can confirm that during testing.

komalsukhani (approved):

PR #687, opened 2024-04-01
[TT-4880] Added support of OAS

komalsukhani:
## Description

Added new CRD to support Tyk OAS.

## Related Issue

[TT-4880](https://tyktech.atlassian.net/browse/TT-4880)

Added end-to-end tests for CRUD operations

## Types of changes

## Checklist

- [x] If you've changed API models, please update CRDs.
  - [x] `make manifests`
  - [x] `make helm`
  - [x] `gofmt -s -w .`
  - [x] `go vet ./...`
  - [x] `golangci-lint run`

[TT-4880]: https://tyktech.atlassian.net/browse/TT-4880?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


controllers/tykoas_controller.go
```diff
@@ -0,0 +1,324 @@
...
+	return r.Client.Status().Update(ctx, tykOASCrd)
+}
+
+func (r *TykOASReconciler) getAPIID(tykOASCrd *v1alpha1.TykOAS, tykOASDoc string) (string, error) {
+	if tykOASCrd.Status.ID != "" {
+		return tykOASCrd.Status.ID, nil
+	}
+
+	val, err := jsonparser.GetString([]byte(tykOASDoc), TykOASExtenstionStr, "info", "id")
+	if err != nil && err != jsonparser.KeyPathNotFoundError {
```
buraksekili, line 252: `KeyPathNotFoundError` is ignored because of that `id` may not exists on all OAS docs, right? Maybe we can add a comment line here or method level to describe this case.

controllers/tykoas_controller.go
```diff
@@ -0,0 +1,324 @@
...
+			util.RemoveFinalizer(tykOAS, keys.TykOASFinalizerName)
+
+			return nil
+		}
+
+		util.AddFinalizer(tykOAS, keys.TykOASFinalizerName)
+
+		apiID, err = r.createOrUpdateTykOASAPI(ctx, tykOAS)
+		if err != nil {
+			r.Log.Error(err, "Failed to create/update Tyk OAS API", "name", client.ObjectKeyFromObject(tykOAS))
```
buraksekili, line 112: I guess we can use `log` here instead of `r.Log` as we have access to `log` 
```suggestion
			log.Error(err, "Failed to create/update Tyk OAS API")
```

controllers/tykoas_controller.go
```diff
@@ -0,0 +1,324 @@
...
+		}
+	}
+
+	return apiID, nil
+}
+
+func (r *TykOASReconciler) updateStatus(ctx context.Context, tykOASCrd *v1alpha1.TykOAS, apiID string) error {
+	var cm v1.ConfigMap
+
+	r.Log.Info("Updating status of Tyk OAS instance")
```
buraksekili, line 194: From looking at this log, it is a bit unclear which OAS CR is being updated at the moment. Again, either we might consider using `log`, or indicating `tykOASCrd`'s metadata information in the log

controllers/tykoas_controller.go
```diff
@@ -0,0 +1,324 @@
...
+
+	val, err := jsonparser.GetString([]byte(tykOASDoc), TykOASExtenstionStr, "info", "id")
+	if err != nil && err != jsonparser.KeyPathNotFoundError {
+		return "", err
+	}
+
+	return val, nil
+}
+
+func (r *TykOASReconciler) delete(ctx context.Context, tykOASCrd *v1alpha1.TykOAS) error {
```
buraksekili, line 259: `delete` only deletes OAS Api from Tyk right? I am asking this because in other controllers, `delete` method also deletes finalziers, so maybe we can either directly use `klient.Universal.TykOAS().Delete()` or rename the method name, something like `deleteFromTyk`

integration/tykoas_test.go
```diff
@@ -0,0 +1,221 @@
...
+			eval.NoErr(err)
+
+			err = waitForTykResourceCreation(c, tykOAS)
+			eval.NoErr(err)
+
+			return ctx
+		}).Assess("Ensure OAS definition is created on Tyk",
+		func(ctx context.Context, t *testing.T, c *envconf.Config) context.Context {
+			exists := klient.Universal.TykOAS().Exists(tykCtx, tykOAS.Status.ID)
+			eval.True(exists)
```
buraksekili, line 68: Maybe we can retry if `exists` variable does not match desired outcome? What do you think? Intermittent connection issues between controller and Tyk might failure on this test case even OAS has been created on Tyk

integration/tykoas_test.go
```diff
@@ -0,0 +1,221 @@
...
+			result, err := jsonparser.Set([]byte(oasDoc), []byte(quotedListenPath),
+				controllers.TykOASExtenstionStr, "server", "listenPath", "value")
+			eval.NoErr(err)
+
+			cm.Data[testOASConfKeyName] = string(result)
+
+			err = c.Client().Resources(testNS).Update(ctx, cm)
+			eval.NoErr(err)
+
+			// Create ApiDefinition Reconciler.
```
buraksekili, line 188: I guess, instead of `ApiDefinition`, it should be `TykOASReconciler`
```suggestion
			// Create TykOASReconciler.
```

integration/tykoas_test.go
```diff
@@ -0,0 +1,221 @@
...
+				Log:    log.NullLogger{},
+				Scheme: cl.Scheme(),
+				Env:    tykEnv,
+			}
+
+			err = wait.For(conditions.New(c.Client().Resources(testNS)).ResourceMatch(tykOAS, func(object k8s.Object) bool {
+				_, err := r.Reconcile(ctx, ctrl.Request{NamespacedName: cr.ObjectKeyFromObject(tykOAS)})
+				return err == nil
+			}))
+			eval.NoErr(err)
```
buraksekili, line 203: Do we need to trigger reconciliation manually? If updating linked configmap triggers reconciliation on OAS CR, i guess we do not need to trigger reconcliation again

integration/tykoas_test.go
```diff
@@ -0,0 +1,221 @@
...
+
+			var tykOASApi string
+
+			tykOASApi, err = klient.Universal.TykOAS().Get(tykCtx, tykOAS.Status.ID)
+			eval.NoErr(err)
```
buraksekili, line 208: I'd suggest to have some retry mechanism here as intermittent errors such as network partition may cause direct failures here.

controllers/tykoas_controller.go
```diff
@@ -0,0 +1,324 @@
...
+func (r *TykOASReconciler) configmapEvents() predicate.Predicate {
+	return predicate.Funcs{
+		CreateFunc: func(e event.CreateEvent) bool {
+			return false
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			return true
+		},
+		DeleteFunc: func(e event.DeleteEvent) bool {
+			return true
```
buraksekili, line 321: what happens to OAS CR when ConfigMap is deleted?

api/v1alpha1/tykoas_types.go
```diff
@@ -0,0 +1,90 @@
...
+type TykOASReference struct {
+	// ConfigmapRef provides information of configmap in which Tyk OAS is stored
+	ConfigmapRef ConfigMapReference `json:"configmapRef"`
+}
+
+type ConfigMapReference struct {
+	// Name is the name of configmap
+	Name string `json:"name"`
+	// Namespace is the namespace of configmap
+	Namespace string `json:"namespace"`
```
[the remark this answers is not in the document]
komalsukhani: We have a separate ticket for it. So we can address it there.

integration/tykoas_test.go
komalsukhani: I tried without explicitly calling Reconcile but it didn't work.

config/samples/oas/petstore.yaml
```diff
@@ -0,0 +1,44 @@
...
+  test_oas.json: |-
+    {
+        "info": {
+          "title": "Petstore",
+          "version": "1.0.0"
+        },
+        "openapi": "3.0.3",
+        "components": {},
+        "paths": {},
+        "x-tyk-api-gateway": {
```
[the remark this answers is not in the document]
komalsukhani: Make sense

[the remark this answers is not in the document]
g-bohncke: I was thinking of building my own operator for this functionality. but I'm glad to see that this is becoming default functionality. The only bit I'm missing/would change that would solve our companies use case is instead of the config map having the OAS definition itself it would have a URL to where to download the definition form. that way when a team create a micro service the want to automatically expose via tyk the don't need a additional manual step of coping their definition in a config map. or having ci/cd do that for them.

A change of the annotation on the TykOAS will be enough to trigger a reload of this all. 

Their might also be a nice gain if the TykOAS would also include an ApiDefinition with the same [template ](template: "true") flags as used for an ingress to make a merge between missing features that can't be defined in OAS but are TYK features.

Sorry if this is not the place for this feature req. if you want it moved i can do so.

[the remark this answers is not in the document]
komalsukhani: Thank you @g-bohncke for such early feedback on the implementation. It always great to use-case of users.
Our plan was to support configMap in MVP and extend in future to support different sources.
It would be great if you raise a Feature Request on Github so we can discuss about implementation there.

> Their might also be a nice gain if the TykOAS would also include an ApiDefinition with the same template flags as used for an ingress to make a merge between missing features that can't be defined in OAS but are TYK features.
In our upcoming 5.3 release, OAS will support majority of Classic API features.

controllers/tykoas_controller.go
```diff
@@ -0,0 +1,324 @@
...
+	var cm v1.ConfigMap
+
+	ns := types.NamespacedName{
+		Name:      tykOASCrd.Spec.TykOAS.ConfigmapRef.Name,
+		Namespace: tykOASCrd.Spec.TykOAS.ConfigmapRef.Namespace,
+	}
+
+	err := r.Client.Get(ctx, ns, &cm)
+	if err != nil {
+		r.Log.Error(err, "Failed to fetch config map", "name", ns)
```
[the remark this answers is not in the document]
komalsukhani: Fixed in [e3d178e](https://github.com/TykTechnologies/tyk-operator/pull/687/commits/e3d178ef016740ad1622181e3b43778641d34e37)

controllers/tykoas_controller.go
komalsukhani: Resolved in [e3d178e](https://github.com/TykTechnologies/tyk-operator/pull/687/commits/e3d178ef016740ad1622181e3b43778641d34e37)

controllers/tykoas_controller.go
komalsukhani: Resolved in [e3d178e](https://github.com/TykTechnologies/tyk-operator/pull/687/commits/e3d178ef016740ad1622181e3b43778641d34e37)

controllers/tykoas_controller.go
komalsukhani: Moved call to remove finalizer inside delete function [e5260e6](https://github.com/TykTechnologies/tyk-operator/pull/687/commits/e5260e6a4c7dc007775f8f8d178f6c6a229075a5)

pkg/client/gateway/tyk_oas.go
```diff
@@ -0,0 +1,111 @@
...
+func (oas TykOAS) Update(ctx context.Context, id, data string) error {
+	quotedID := fmt.Sprintf("\"%s\"", id)
+
+	// Need to set API ID in the body too for update to work
+	result, err := jsonparser.Set([]byte(data), []byte(quotedID), "x-tyk-api-gateway", "info", "id")
+	if err != nil {
+		return err
+	}
+
+	resp, err := client.Put(ctx, client.Join(endpointOAS, id), bytes.NewBuffer(result))
```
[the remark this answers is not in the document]
komalsukhani: PUT is used to update the Tyk OAS API. PATCH is used if user has updated OAS Document and want Tyk to update their API Definition accordingly.

In our case, we only support Tyk OAS API. So we need to use PUT.

integration/tykoas_test.go
komalsukhani: Resolved in [66cb5ce](https://github.com/TykTechnologies/tyk-operator/pull/687/commits/66cb5cef364d99ac5601a241c0a21b8e4242dde0)

integration/tykoas_test.go
komalsukhani: Resolved in [532edf1](https://github.com/TykTechnologies/tyk-operator/pull/687/commits/532edf15773abe0e351dd3bac59da6b6e1ad551b)

buraksekili (approved):

PR #705, opened 2024-05-14
[TT-8086] Add support of client certificates in OAS

komalsukhani:
## Description

Add support of client certificates in OAS

## Related Issue

[TT-8086](https://tyktech.atlassian.net/browse/TT-8086)

## Types of changes

## Checklist

- [x] If you've changed API models, please update CRDs.
  - [x] `make manifests`
  - [x] `make helm`
  - [x] `gofmt -s -w .`
  - [x] `go vet ./...`
  - [x] `golangci-lint run`

[TT-8086]: https://tyktech.atlassian.net/browse/TT-8086?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


controllers/secret_cert_controller.go
```diff
@@ -177,31 +275,24 @@ func (r *SecretCertReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
 			if !isCertificateAlreadyUploaded(ctx, isCertPreviouslyProcessed, tlsCrt, env.Org) {
-				certID, err = klient.Universal.Certificate().Upload(ctx, tlsKey, tlsCrt)
+				certID, err := klient.Universal.Certificate().Upload(ctx, tlsKey, tlsCrt)
 				if err != nil {
-					return ctrl.Result{Requeue: true}, err
+					return err
 				}
 
-				log.Info("uploaded certificate to Tyk", "certID", certID)
+				log.Info("Successfully uploaded certificate to Tykk", "certID", certID)
```
olamilekan000, line 295: you have a little typo here. :)

config/samples/oas/client_certificate.yaml
```diff
@@ -0,0 +1,51 @@
...
+kind: TykOasApiDefinition
+metadata:
+  name: petstore
+spec:
+  tykOAS:
+    configmapRef:
+      name: cm
+      namespace: default
+      keyName: test_oas.json
+  client_certificate_ref: [tls-cert]
```
buraksekili, line 51: I guess this should be camelCase as it is defined as camelCase in Go spec.
```suggestion
  clientCertificateRef: [tls-cert]
```

controllers/oas_parser.go
```diff
@@ -0,0 +1,41 @@
...
+	val, _, _, err := jsonparser.Get(data, ClientCertificatePath...)
+	if err != nil && err != jsonparser.KeyPathNotFoundError {
+		return nil, err
+	}
+
+	if val != nil {
+		_, err = jsonparser.ArrayEach(val, func(value []byte, dataType jsonparser.ValueType, offset int, err error) {
+			item, pErr := jsonparser.ParseString(value)
+			if pErr != nil {
+				return
```
buraksekili, line 23: if `pErr` is not nil, we return `err` not `pErr`. Is this behaviour expected? I assume we need to return pErr unless I miss some specific implementation details. Here, do we want to continue parsing certificates array if the parser is not able to parse the field within the array?

controllers/oas_parser.go
```diff
@@ -0,0 +1,41 @@
...
+var ClientCertificatePath = []string{TykOASExtenstionStr, "server", "clientCertificates", "allowlist"}
+
+func OASSetClientCertificates(data []byte, certid string) ([]byte, error) {
+	allowlist := []string{}
+
+	val, _, _, err := jsonparser.Get(data, ClientCertificatePath...)
+	if err != nil && err != jsonparser.KeyPathNotFoundError {
+		return nil, err
+	}
+
```
buraksekili, line 18: let's move the creation of `allowList` below the following `if` condition. If the `if` statement fails, we don't need `allowList`, and creating it would be unnecessary.

```suggestion
	val, _, _, err := jsonparser.Get(data, ClientCertificatePath...)
	if err != nil && err != jsonparser.KeyPathNotFoundError {
		return nil, err
	}
	
	allowlist := []string{}

```

controllers/secret_cert_controller.go
```diff
@@ -100,36 +102,139 @@ func (r *SecretCertReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+) error {
+	tykOasList := v1alpha1.TykOasApiDefinitionList{}
+
+	opts := []client.ListOption{
+		client.InNamespace(secretNS.Namespace),
+	}
+
+	// list all the Tyk OAS Apis in the namespace of the secret
+	if err := r.List(ctx, &tykOasList, opts...); err != nil {
+		log.Info("unable to list api definitions")
```
buraksekili, line 130:
```suggestion
		log.Info("failed to list TykOasApiDefinitions", "namespace", secretNS.Namespace)
```

controllers/tykoasapidefinition_controller.go
```diff
@@ -175,22 +176,70 @@ func (r *TykOasApiDefinitionReconciler) createOrUpdateTykOASAPI(ctx context.Cont
 		apiID = EncodeNS(client.ObjectKeyFromObject(tykOASCrd).String())
 	}
 
+	if tykOASCrd.Spec.ClientCertificateRef != nil {
+		if tykOASDoc, err = r.processClientCertificateRef(ctx, env, log, tykOASCrd, tykOASDoc); err != nil {
+			log.Error(err, "failed to process client certificate")
```
buraksekili, line 181: do we want to print this message as returning errors cause operator manager to print out the `err` as well?

integration/secret_cert_test.go
```diff
@@ -127,3 +129,130 @@ func TestCertificateUpload(t *testing.T) {
...
+			eval := is.New(t)
+			var err error
+
+			testNS, _ = ctx.Value(ctxNSKey).(string) // nolint: errcheck
+
+			// Obtain Environment configuration to be able to connect Tyk.
+			tykEnv, err = generateEnvConfig(ctx, envConf)
+			eval.NoErr(err)
+
+			res := semver.Compare(tykEnv.TykVersion, "v4.1")
```
buraksekili, line 153: we'll we use 4.1 or 5.3 as the starting point for OAS. based on that we should align our code-base because we have discrepancy in oas versions now

integration/secret_cert_test.go
```diff
@@ -127,3 +129,130 @@ func TestCertificateUpload(t *testing.T) {
...
+					  "strip": true
+					},
+					"clientCertificates":{
+					  "enabled": true
+					}
+				  }
+				}
+			  }`
+
+			// Create test api definition
```
buraksekili, line 199:
```suggestion
			// Create test TykOasApiDefinition
```

integration/secret_cert_test.go
```diff
@@ -127,3 +129,130 @@ func TestCertificateUpload(t *testing.T) {
...
+				  "upstream": {
+					"url": "https://petstore.swagger.io/v2"
+				  },
+				  "server": {
+					"listenPath": {
+					  "value": "/petstore/",
+					  "strip": true
+					},
+					"clientCertificates":{
+					  "enabled": true
```
buraksekili, line 193: Does `clientCertificates` field needs to be set? What happens if users specify `clientCertificateRef` field in CRD while the OAS JSON file in ConfigMap lacks of 
```json
"clientCertificates":{"enabled": true}
```

I now that enabled needs to be set to true explicitly but i am wondering what happens if there is no `clientCertificates` field in OAS ConfigMap

integration/secret_cert_test.go
```diff
@@ -127,3 +129,130 @@ func TestCertificateUpload(t *testing.T) {
...
+		func(ctx context.Context, t *testing.T, envConf *envconf.Config) context.Context {
+			eval := is.New(t)
+
+			// Validate the certificate is uploaded
+			err := wait.For(func() (done bool, err error) {
+				tykOASApi, err := klient.Universal.TykOAS().Get(tykCtx, tykOas.Status.ID)
+				if err != nil {
+					t.Log("Failed to fetch Tyk OAS API", "id", tykOas.Status.ID, "error", err)
+
+					return false, err
```
buraksekili, line 222: Returning an err will halt the execution. So, sometimes - especially in OSS gateway - creation of files takes a while. I'd suggest retry getting the OAS until timeout occurs

controllers/tykoasapidefinition_controller.go
komalsukhani: I would like to keep it as it gives context of which operation failed because of the error.

integration/secret_cert_test.go
komalsukhani: Makes sense

controllers/oas_parser.go
komalsukhani: The function `ArrayEach` accepts doesn't return an error. So I can't return pErr from it.
I am checking error returned by ArrayEach itself.

integration/secret_cert_test.go
komalsukhani: It is still a valid API Definition. We support MTLS via 2 flows. 
1. static MTLS: Here is certificate is provided in the API Definition itself
2. dynamic MTLS: User can provide certificate in the key, upload keys through developer portal 
Reference to docs https://tyk.io/docs/basic-config-and-security/security/mutual-tls/client-mtls/

controllers/oas_parser.go
buraksekili: oh my bad, you are right

integration/secret_cert_test.go
buraksekili: Yes, that will be valid but i would like to know whether operator will be able to parse and populate non-existing field in the JSON object.

As far as I am concerned, `OASSetClientCertificates` handles non existing keys but would like to double check with you

buraksekili (approved):

integration/secret_cert_test.go
komalsukhani: It handles it.

buraksekili (approved):

PR #709, opened 2024-05-27
TT-11998: add features section to graphql proxy config

pvormste:
This PR adds a new `features` section to the graphql proxy-only config inside the API definition.

## Types of changes


buraksezer (approved):

api/model/api.go
```diff
@@ -1200,6 +1200,8 @@ type GraphQLResponseExtensions struct {
 }
 
 type GraphQLProxyConfig struct {
+	// +nullable
+	Features GraphQLProxyFeaturesConfig `json:"features"`
```
buraksekili, line 1204: @pvormste will this feature be required for users to provide? If not, we can make it optional as follows:
```suggestion
	Features GraphQLProxyFeaturesConfig `json:"features,omitempty"`
```

api/model/api.go
```diff
@@ -1210,6 +1212,10 @@ type GraphQLProxyConfig struct {
 	RequestHeadersRewrite map[string]RequestHeadersRewriteConfig `json:"request_headers_rewrite,omitempty"`
 }
 
+type GraphQLProxyFeaturesConfig struct {
+	UseImmutableHeaders bool `json:"use_immutable_headers"`
```
buraksekili, line 1216: Similar to my comment above, if the feature is not required to be specified, let's make it optional.

buraksekili (approved):

PR #719, opened 2024-06-27
TT-11509: manage OAS API Versions using the Operator

olamilekan000:
## Description
```
change manages OAS API versions using the Operator.
```

https://tyktech.atlassian.net/browse/TT-11509
https://tyktech.atlassian.net/browse/TT-12602

## Types of changes

## Checklist

  - [ ] `make manifests`
  - [ ] `make helm`
  - [ ] `gofmt -s -w .`
  - [ ] `go vet ./...`
  - [ ] `golangci-lint run`


api/v1alpha1/tykoasapidefinition_types.go
```diff
@@ -48,6 +52,38 @@ type TykOASReference struct {
...
+
+	// Versions contains a list of versions that map to individual API IDs.
+	Versions []TykOASVersions `json:"versions,omitempty"`
+}
+
+type TykOASVersions struct {
+	// Name contains the name of the refrenced OasApiDefinition.
+	Name string `json:"name"`
+
+	// OasApiDefinitionRef refrences an OasApiDefinition.
```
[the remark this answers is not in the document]
komalsukhani: I would suggest to use `model.Target` type for this field here to have consistency with other CRD design

api/v1alpha1/tykoasapidefinition_types.go
```diff
@@ -75,6 +111,14 @@ type TykOasApiDefinitionStatus struct {
 	LatestTransaction TransactionInfo `json:"latestTransaction,omitempty"`
 	// IngressTemplate shows whether this CR is used as Ingress Template or not.
 	IngressTemplate bool `json:"ingressTemplate,omitempty"`
+	// IsVersionedAPI indicates if the API is a version or not.
+	IsVersionedAPI bool `json:"isVersionedAPI,omitempty"`
+	// BaseVersionName specifies the name of the Base OAS API if IsVersionedAPI is set to true.
+	BaseVersionName string `json:"baseVersionName,omitempty"`
```
komalsukhani, line 117: I would should to use `model.Target` type to store both name and namespace of base type.

controllers/tykoasapidefinition_controller.go
```diff
@@ -188,6 +220,173 @@ func (r *TykOasApiDefinitionReconciler) Reconcile(ctx context.Context, req ctrl.
...
+		return err
+	}
+
+	byteData := []byte(tykBaseOASDoc)
+
+	if byteData, err = jsonparser.Set(byteData, byteVersioning, "x-tyk-api-gateway", "info", "versioning"); err != nil {
+		return err
+	}
+
+	if err := klient.Universal.TykOAS().Update(ctx, baseAPIID, string(byteData)); err != nil {
```
komalsukhani, line 381: Instead of calling Update function here, just update the `tykBaseOASDoc` and let `createOrUpdateTykOASAPI` call update/create method.
This will reduce number of calls we are making to Tyk.

controllers/tykoasapidefinition_controller.go
```diff
@@ -188,6 +220,173 @@ func (r *TykOasApiDefinitionReconciler) Reconcile(ctx context.Context, req ctrl.
...
+			Namespace: versionedTykOAS.Status.BaseVersionNamespace,
+		}, &baseOasAPI); err != nil {
+			if k8serrors.IsNotFound(err) {
+				log.Info("failed to get TykOasApiDefinition CR", "err", err.Error())
+			}
+
+			return err
+		}
+
+		errMsg := "Cannot delete API. update the Base OasAPI to delete"
```
komalsukhani, line 257:
```suggestion
		errMsg := "Cannot delete API. It is referenced as version in another Tyk OAS API. Please remove it before deleting."
```

controllers/tykoasapidefinition_controller.go
```diff
@@ -188,6 +220,173 @@ func (r *TykOasApiDefinitionReconciler) Reconcile(ctx context.Context, req ctrl.
...
 }
 
+func (r *TykOasApiDefinitionReconciler) handleVersionedTykOASAPI(
+	ctx context.Context,
+	log logr.Logger,
+	versionedTykOAS *v1alpha1.TykOasApiDefinition,
+) error {
+	if !versionedTykOAS.ObjectMeta.DeletionTimestamp.IsZero() {
+		if versionedTykOAS.Status.IsDefaultVersion {
+			errMsg := "Cannot delete default API version. Remove it from base API's default first"
```
komalsukhani, line 230: Do log this error message.

controllers/tykoasapidefinition_controller.go
komalsukhani: Do log this error message

controllers/tykoasapidefinition_controller.go
```diff
@@ -188,6 +220,173 @@ func (r *TykOasApiDefinitionReconciler) Reconcile(ctx context.Context, req ctrl.
...
+			}
+
+			return err
+		}
+
+		var oasAPIPresent bool
+
+		for _, v := range baseOasAPI.Spec.Versioning.Versions {
+			if v.OasApiDefinitionRef == versionedTykOAS.Name {
+				oasAPIPresent = true
```
komalsukhani, line 301:
```suggestion
				oasAPIPresent = true
				break
```

controllers/tykoasapidefinition_controller.go
```diff
@@ -411,6 +622,41 @@ func deleteOasApi(ctx context.Context, tykOASCrd *v1alpha1.TykOasApiDefinition)
 	return nil
 }
 
+func (r *TykOasApiDefinitionReconciler) updateBaseOasAPIVersionsStatus(
```
komalsukhani, line 625:
```suggestion
func (r *TykOasApiDefinitionReconciler) updateOasAPIVersionsStatus(
```

controllers/tykoasapidefinition_controller.go
```diff
@@ -188,6 +220,173 @@ func (r *TykOasApiDefinitionReconciler) Reconcile(ctx context.Context, req ctrl.
 	return ctrl.Result{RequeueAfter: reqA}, err
 }
 
+func (r *TykOasApiDefinitionReconciler) handleVersionedTykOASAPI(
```
komalsukhani, line 223: Can you add a comment above function explaining what it does?

controllers/tykoasapidefinition_controller.go
```diff
@@ -188,6 +220,173 @@ func (r *TykOasApiDefinitionReconciler) Reconcile(ctx context.Context, req ctrl.
...
+			}
+		}
+
+		return nil
+	}
+
+	return nil
+}
+
+func (r *TykOasApiDefinitionReconciler) handleVersioningTykOASAPI(ctx context.Context,
```
komalsukhani, line 319: Can you add comment above the function?

komalsukhani (changes requested):

komalsukhani:
@olamilekan000 Can you add integration(can be added in integration folder)/unit tests?
Also, `handleVersionedTykOASAPI` and `handleVersioningTykOASAPI` names are confusing. Can you name it different?

api/v1alpha1/tykoasapidefinition_types.go
```diff
@@ -48,6 +52,38 @@ type TykOASReference struct {
...
+
+	// Enabled is a boolean flag, if set to true it will enable versioning of the API.
+	Enabled bool `json:"enabled"`
+
+	// Key contains the name of the key to check for versioning information.
+	Key string `json:"key"`
+
+	// Location contains versioning location information. It can be one of the following:
+	// header, url-param, url.
+	Location string `json:"location"`
```
buraksekili, line 67: if `location` only accepts `header`, `url-param` or `url`, it is worth to add some validation here. Please see: https://book.kubebuilder.io/reference/markers/crd-validation#crd-validation

controllers/tykoasapidefinition_controller.go
```diff
@@ -188,6 +220,173 @@ func (r *TykOasApiDefinitionReconciler) Reconcile(ctx context.Context, req ctrl.
...
+	for _, v := range tykBaseOASCrd.Spec.Versioning.Versions {
+		var versionOasAPI v1alpha1.TykOasApiDefinition
+
+		if err := r.Get(ctx, types.NamespacedName{
+			Name:      v.OasApiDefinitionRef,
+			Namespace: v.Namespace,
+		}, &versionOasAPI); err != nil {
+			if k8serrors.IsNotFound(err) {
+				log.Info("failed to get versioned TykOasApiDefinition CR", "err", err.Error())
+			}
```
buraksekili, line 336: why do we only print the log when the resource is not found on k8s? in my opinion, as we return from the loop whenever an error occurs, it'd be better to print the log whenever an error occurs. what do you think?

api/v1alpha1/tykoasapidefinition_types.go
```diff
@@ -75,6 +111,14 @@ type TykOasApiDefinitionStatus struct {
...
 	// IngressTemplate shows whether this CR is used as Ingress Template or not.
 	IngressTemplate bool `json:"ingressTemplate,omitempty"`
+	// IsVersionedAPI indicates if the API is a version or not.
+	IsVersionedAPI bool `json:"isVersionedAPI,omitempty"`
+	// BaseVersionName specifies the name of the Base OAS API if IsVersionedAPI is set to true.
+	BaseVersionName string `json:"baseVersionName,omitempty"`
+	// BaseVersionNamespace specifies the namespace of the Base OAS API.
+	BaseVersionNamespace string `json:"baseVersionNamespace,omitempty"`
+	// IsDefaultVersion specifies if the OAS API is the default  Version.
+	IsDefaultVersion bool `json:"isDefaultVersion,omitempty"`
```
buraksekili, line 121: i would suggest moving all versioning-related info to a new sub-structure maybe called as `*VersioningStatus`. And set it to nil if the OAS API is not versioned.

api/v1alpha1/tykoasapidefinition_types.go
```diff
@@ -95,6 +139,13 @@ type TykOasApiDefinition struct {
 	Status TykOasApiDefinitionStatus `json:"status,omitempty"`
 }
 
+func (t *TykOasApiDefinition) RemoveOASVersionStatus() {
```
buraksekili, line 142: instead of setting each field to its empty values, i'd suggest setting the sub-struct to nil - so that in the future if we add/remove fields regarding versioning we won't forget to add a logic to delete them in this method.
For example, if all versioning-related information is stored in a struct, let's say `*VersioningStatus `, setting `t.Status.VersioningStatus = nil` is simpler

controllers/tykoasapidefinition_controller.go
```diff
@@ -75,6 +76,21 @@ type TykOasApiDefinitionReconciler struct {
 	Scheme *runtime.Scheme
 }
 
+type TykOasApiDefinitionVersion struct {
```
buraksekili, line 79: please add a comment about the struct, why, and when it is being used.

controllers/tykoasapidefinition_controller.go
```diff
@@ -75,6 +76,21 @@ type TykOasApiDefinitionReconciler struct {
...
+	ID   string `json:"id"`
+	Name string `json:"name"`
+}
+
+// TykOasApiDefinitionVersion represents the version information for a base TykOasApiDefinition object
```
buraksekili, line 84:
```suggestion
// TykOasApiDefinitionVersioning represents the version information for a base TykOasApiDefinition object
```

buraksekili: also, please add comments about the purpose of the structure.

controllers/tykoasapidefinition_controller.go
```diff
@@ -138,10 +154,25 @@ func (r *TykOasApiDefinitionReconciler) Reconcile(ctx context.Context, req ctrl.
 		return ctrl.Result{}, nil
 	}
 
+	if tykOAS.Status.IsVersionedAPI {
+		err := r.handleVersionedTykOASAPI(ctx, log, tykOAS)
```
buraksekili, line 158: what do you think renaming this method name (`handleVersionedTykOASAPI`) to align its name with other methods; for instance, `reconcileVersionedOasApi` or sth like this.

controllers/tykoasapidefinition_controller.go
```diff
@@ -138,10 +154,25 @@ func (r *TykOasApiDefinitionReconciler) Reconcile(ctx context.Context, req ctrl.
...
+		}
+
+		return ctrl.Result{}, nil
+	}
+
 	_, err = util.CreateOrUpdate(ctx, r.Client, tykOAS, func() error {
 		if !tykOAS.ObjectMeta.DeletionTimestamp.IsZero() {
 			markForDeletion = true
 
+			if err := r.updateBaseOasAPIVersionsStatus(ctx, log, tykOAS); err != nil {
```
buraksekili, line 172: As far as i am concerned, this method only runs during the reconciliation to update the status of base API, right?. If that's the case, i'd suggest renaming the method name because the method must only update the status only during deletion. So, it'd be better to rename it like `reconcileDeletingBase` sth sth.

controllers/tykoasapidefinition_controller.go
```diff
@@ -188,6 +220,173 @@ func (r *TykOasApiDefinitionReconciler) Reconcile(ctx context.Context, req ctrl.
...
+		versionOasAPI.Status.IsVersionedAPI = true
+
+		version := TykOasApiDefinitionVersion{
+			ID:   versionOasAPI.Status.ID,
+			Name: v.Name,
+		}
+
+		versions = append(versions, &version)
+
+		if err := r.Status().Update(ctx, &versionOasAPI); err != nil {
```
buraksekili, line 356: updating status here is tricky because this method is called within CreateOrUpdate which reconciles resources within its closure function. and updating the status would cause conflicts on k8s side - which might lead redundant reconciliations to recover.

controllers/tykoasapidefinition_controller.go
```diff
@@ -188,6 +220,173 @@ func (r *TykOasApiDefinitionReconciler) Reconcile(ctx context.Context, req ctrl.
...
+	}
+
+	byteVersioning, err := json.Marshal(versioning)
+	if err != nil {
+		return err
+	}
+
+	byteData := []byte(tykBaseOASDoc)
+
+	if byteData, err = jsonparser.Set(byteData, byteVersioning, "x-tyk-api-gateway", "info", "versioning"); err != nil {
```
buraksekili, line 377: I would suggest moving the fields used here (`"x-tyk-api-gateway", "info", "versioning"`) to global section so that they can be used later, maybe in tests. For reference, https://github.com/TykTechnologies/tyk-operator/blob/ed7b42c07978a57229219e2711be6bad76c43231/controllers/tykoasapidefinition_controller.go#L58-L69

controllers/tykoasapidefinition_controller.go
```diff
@@ -411,6 +622,41 @@ func deleteOasApi(ctx context.Context, tykOASCrd *v1alpha1.TykOasApiDefinition)
...
+			if err := r.Get(ctx, types.NamespacedName{
+				Name:      v.OasApiDefinitionRef,
+				Namespace: v.Namespace,
+			}, &versionOasAPI); err != nil {
+				if k8serrors.IsNotFound(err) {
+					log.Info("failed to get versioned TykOasApiDefinition CR", "err", err.Error())
+				}
+
+				return err
+			}
```
buraksekili, line 647: Please correct me if i am wrong. This logic only runs while reconciling the delete request and I couldn't get why non-existent OAS API Definitions on k8s prevents this resource from deletion?

controllers/tykoasapidefinition_controller.go
```diff
@@ -411,6 +622,41 @@ func deleteOasApi(ctx context.Context, tykOASCrd *v1alpha1.TykOasApiDefinition)
...
+
+			versionOasAPI.RemoveOASVersionStatus()
+
+			if err := r.Status().Update(ctx, &versionOasAPI); err != nil {
```
buraksekili, line 651: Again, updating the status here might lead to conflicts. please check the logs if that's the case - if so, please update the flow of the reconciliation to prevent such conflicts.

controllers/tykoasapidefinition_controller.go
olamilekan000: This type of update needs to happen when the baseAPI is marked for deletion. What is happening in the function is that, all versions are being detached from the base API when a delete of the base API is triggered.

controllers/tykoasapidefinition_controller.go
olamilekan000: what do you mean by conflict? can you give more context?

controllers/tykoasapidefinition_controller.go
buraksekili: i see why it is needed. the problem is related how kubernetes api handles updates. this function called inside CreateOrUpdate, [here](https://github.com/TykTechnologies/tyk-operator/blob/153761130cf0fbb95a9a2f3b3b010b57eea9203a/controllers/tykoasapidefinition_controller.go#L183).
If you check [the docs](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.9.0/pkg/controller/controllerutil#CreateOrUpdate) of CreateOrUpdate, you will see that the CreateOrUpdate function fetches the resource from kubernetes, then runs the function that we passed into that and Updates the resource. In the meantime, if you update resource's Status field, there might be a drift in `resourceVersion` of the resource, which will cause conflicts on kubernetes side.
Let me try to simplify, assume that you have a CRD called `X` which has `resourceVersion: 1`.
```
function CreateOrUpdate starts
        fetches X from API server
        sees X is having a resourceVersion: 1
        runs mutate function that we passed into CreateOrUpdate
                ---> mutate function starts
                the mutate function updates status which will update the resourceVersion 
                    of the object and new resourceVersion will be different than 1.
               ---> mutate function finishes
        CreateOrUpdate sends the object to the Api Server to update it.
        It will fail because the object that we reconciled within API Server 
              has resourceVersion: 1 while the API server has the 
              same object with different resourceVersion. 
              That's why the manager will requeue the object again 
              and this loop might stuck. 
```

controllers/tykoasapidefinition_controller.go
buraksekili: please refer to this comment: https://github.com/TykTechnologies/tyk-operator/pull/719#discussion_r1661993294

api/v1alpha1/tykoasapidefinition_types.go
```diff
@@ -48,6 +52,39 @@ type TykOASReference struct {
...
+
+	// Enabled is a boolean flag, if set to true it will enable versioning of the API.
+	Enabled bool `json:"enabled"`
+
+	// Key contains the name of the key to check for versioning information.
+	Key string `json:"key"`
+
+	// Location contains versioning location information. It can be one of the following:
+	// header, url-param, url.
+	// +kubebuilder:validation:Enum=header;url-param;url
```
buraksekili, line 78: it'd be better if you move this validation enum to the where `LocationType` is defined above.

controllers/tykoasapidefinition_controller.go
```diff
@@ -188,6 +224,192 @@ func (r *TykOasApiDefinitionReconciler) Reconcile(ctx context.Context, req ctrl.
...
+		return err
+	}
+
+	if !versionedTykOAS.ObjectMeta.DeletionTimestamp.IsZero() {
+		if versionedTykOAS.Status.GetIsDefaultVersion() {
+			errMsg := "Cannot delete default API version. Remove it from base API's default first"
+
+			versionedTykOAS.Status.LatestTransaction = v1alpha1.TransactionInfo{
+				Time:   metav1.Now(),
+				Status: v1alpha1.TransactionStatus(errMsg),
```
buraksekili, line 244: Status must be [`Failed`](https://github.com/TykTechnologies/tyk-operator/blob/5763498f755e8b8f1a8283616a53ae4dc183f2e3/api/v1alpha1/common.go#L5-L23), whereas the error message can be specified in `Error` field of TransactionInfo struct, [here](https://github.com/TykTechnologies/tyk-operator/blob/5763498f755e8b8f1a8283616a53ae4dc183f2e3/api/v1alpha1/common.go#L33-L34).

controllers/tykoasapidefinition_controller.go
```diff
@@ -188,6 +224,192 @@ func (r *TykOasApiDefinitionReconciler) Reconcile(ctx context.Context, req ctrl.
...
+			}
+
+			log.Info("error deleting versioned TykOasAPI", errMsg)
```
buraksekili, line 247: What do you think updating this log to Error? Are there any specific reason to print Info log here instead of Error?

controllers/tykoasapidefinition_controller.go
buraksekili: @olamilekan000 i still see this is the case. can you please have a look once again? What do you think about this?

controllers/tykoasapidefinition_controller.go
```diff
@@ -188,6 +224,192 @@ func (r *TykOasApiDefinitionReconciler) Reconcile(ctx context.Context, req ctrl.
...
+) error {
+	if versionedTykOAS.Status.GetBaseVersionName() == "" {
+		return nil
+	}
+
+	var baseOasAPI v1alpha1.TykOasApiDefinition
+
+	ns := versionedTykOAS.Status.GetBaseVersionNamespace()
+	if ns == "" {
+		ns = "default"
```
buraksekili, line 299: any specific reason to use `"default"` namespace as a default value? should we use the namespace of `versionedTykOAS`? What do you think?

buraksekili (commented):
Please also update helm files with; `make helm` command

api/v1alpha1/tykoasapidefinition_types.go
olamilekan000: Why should it be on the `LocationType` and not `Location`?

buraksekili: If we add validation to `LocationType` type, whenever we use LocationType, this validation will apply. On the other, with the current approach, the validation only applies the struct field defined here.

controllers/tykoasapidefinition_controller.go
olamilekan000: If i get you correctly, in this function the x will be the baseAPI. However, it is not the baseAPI that is being updated, but the versions and this update happens only when the base API has been marked for deletion.

buraksekili: yes, you are right - mb. please ignore the comment

controllers/tykoasapidefinition_controller.go
olamilekan000:
```		
if err := r.Get(ctx, types.NamespacedName{
			Name:      v.TykOasApiDefinitionRef,
			Namespace: v.Namespace,
		}, &versionOasAPI); err != nil {
			if k8serrors.IsNotFound(err) {
				log.Info("failed to get versioned TykOasApiDefinition CR", "err", err.Error())
			}

			return err
		}
```

This is what the logic looks like, the log is just logging the specific error

controllers/tykoasapidefinition_controller.go
olamilekan000: The reason for `log.Info` is that the error isn't a critical one, it more of a validation type of error. The error log often looks like something critical happened.

api/v1alpha1/tykoasapidefinition_types.go
olamilekan000: it's still the same. When this is converted to CRD, you get something like this. Which is what is expected.

```
      location:
        description: 'Location contains versioning location information.
          It can be one of the following: header, url-param, url.'
        enum:
        - header
        - url-param
        - url
        type: string
```

controllers/tykoasapidefinition_controller.go
buraksekili: my question is why are we logging the specific error? can't we just print all errors?

api/v1alpha1/tykoasapidefinition_types.go
buraksekili: if you think so, lets keep it as is. i'd still suggest moving all validation markers to 1 place instead of scattering them across the code.

controllers/tykoasapidefinition_controller.go
buraksekili: also, we are returning this error, so please use `log.Error`

olamilekan000: https://github.com/TykTechnologies/tyk-operator/pull/719/files#diff-6cd635972dc870184cc2eb7fd04f8ccf46d6821cbec2f194fb419469cc4ba437R456

it's used here already

buraksekili: if it is printed, no need to re-print here as well. please ignore this log message

controllers/tykoasapidefinition_controller.go
```diff
@@ -77,6 +79,22 @@ type TykOasApiDefinitionReconciler struct {
 	Scheme *runtime.Scheme
 }
 
+// TykOasApiDefinitionVersion refrences a version of the Base TykOasAPI.
```
komalsukhani, line 82:
```suggestion
// TykOasApiDefinitionVersion references a version of the Base TykOasAPI.
```

controllers/tykoasapidefinition_controller.go
komalsukhani: But user is trying to delete a CR and it is not getting deleted. Shouldn't it logged as error as operation is unsuccessful?

controllers/tykoasapidefinition_controller.go
```diff
@@ -196,6 +232,182 @@ func (r *TykOasApiDefinitionReconciler) Reconcile(ctx context.Context, req ctrl.
...
+			}
+
+			return err
+		}
+
+		errMsg := "Cannot delete API. It is referenced as version in another Tyk OAS API. Please remove it before deleting."
+
+		versionedTykOAS.Status.LatestTransaction = v1alpha1.TransactionInfo{
+			Time:   metav1.Now(),
+			Status: v1alpha1.TransactionStatus(errMsg),
```
komalsukhani, line 281: Same as what Burak mentioned above https://github.com/TykTechnologies/tyk-operator/pull/719/files#r1662000696

controllers/tykoasapidefinition_controller.go
```diff
@@ -449,6 +670,62 @@ func getAPIID(tykOASCrd *v1alpha1.TykOasApiDefinition, tykOASDoc string) (string
...
+			var versionOasAPI v1alpha1.TykOasApiDefinition
+
+			if err := r.Get(ctx, types.NamespacedName{
+				Name:      v.TykOasApiDefinitionRef,
+				Namespace: v.Namespace,
+			}, &versionOasAPI); err != nil {
+				if k8serrors.IsNotFound(err) {
+					log.Info("failed to get versioned TykOasApiDefinition CR", "err", err.Error())
+
+					return nil
```
komalsukhani, line 712: Instead of returning you should continue to remove reference from versions.

controllers/tykoasapidefinition_controller.go
```diff
@@ -196,6 +232,182 @@ func (r *TykOasApiDefinitionReconciler) Reconcile(ctx context.Context, req ctrl.
...
+		}
+
+		versions = append(versions, &version)
+
+		if err := r.Status().Update(ctx, &versionOasAPI); err != nil {
+			return err
+		}
+	}
+
+	versioning := TykOasApiDefinitionVersioning{
```
komalsukhani, line 386: If I am not wrong, we don't to provide versioning information in `versioned` APIs. So if `versions` field is empty don't pass this information.

controllers/tykoasapidefinition_controller.go
```diff
@@ -196,6 +232,182 @@ func (r *TykOasApiDefinitionReconciler) Reconcile(ctx context.Context, req ctrl.
...
+		if tykBaseOASCrd.Spec.Versioning.Default == v.Name {
+			versionOasAPI.Status.VersioningStatus.IsDefaultVersion = true
+		}
+
+		versionOasAPI.Status.SetBaseVersionName(tykBaseOASCrd.Name)
+		versionOasAPI.Status.SetBaseVersionNamespace(&tykBaseOASCrd.Namespace)
+		versionOasAPI.Status.SetIsVersionedAPI(true)
+
+		version := TykOasApiDefinitionVersion{
+			ID:   versionOasAPI.Status.ID,
```
komalsukhani, line 375: What if versioned API is not created yet and `ID` is empty?

integration/tykoasapidefinition_test.go
```diff
@@ -297,6 +297,121 @@ func TestOASUpdate(t *testing.T) {
...
+					eval.True(ok)
+
+					t.Logf("looking for %+v", currTykOas.Status)
+					return klient.Universal.TykOAS().Exists(tykCtx, currTykOas.Status.ID) == true &&
+						currTykOas.Spec.Versioning.Enabled
+				}))
+			eval.NoErr(err)
+
+			return ctx
+		}).Assess("Test versioned API is default version",
```
komalsukhani, line 396: Can you also check here that status of  version API `tykOAS` is updated correctly?

komalsukhani (changes requested):

controllers/tykoasapidefinition_controller.go
olamilekan000: If it isn't created, it won't get to there. It will get stop at this [line](https://github.com/TykTechnologies/tyk-operator/pull/719/files#diff-6cd635972dc870184cc2eb7fd04f8ccf46d6821cbec2f194fb419469cc4ba437R359)

controllers/tykoasapidefinition_controller.go
olamilekan000: if an error ocuurs, it means that the api version won't be fetched, It will lead to an error if we proceed.

komalsukhani: I mean try to fetch next version.
As we are returning nil, the link between base API and other versions will never be deleted.

controllers/tykoasapidefinition_controller.go
komalsukhani: Can you share line again? It's pointing to empty line right now.

olamilekan000: [here](https://github.com/TykTechnologies/tyk-operator/pull/719/files#diff-6cd635972dc870184cc2eb7fd04f8ccf46d6821cbec2f194fb419469cc4ba437R357)

komalsukhani: ID field in status is only set when API is created on Tyk.
So CRD may exists but it might not be created on Tyk yet.

komalsukhani (approved):

komalsukhani (approved):

bojank93 (approved):

